PR #2, opened 2015-11-03
Change file path patterns to support 1-level symlinks by default

rlhh:
As proposed in the PR for the knapsack repo.

https://github.com/ArturT/knapsack/pull/27


ArturT:
:+1:

PR #4, opened 2016-03-18
Changed rake task in minitest_runner.rb to have no warnings output

patrickbizea:
Due to the changes in [this PR on Rake itself](https://github.com/ruby/rake/pull/97), executing the minitest rake task for Knapsack caused our test output to be filled with so many extraneous warning messages during test execution it both caused the output to be unreadably long and caused the semaphoreci threads to error out, apparently due to lack of memory. This reverts the setting for warnings on the Rake task defined in minitest_runner.rb to the pre-Rake v11.0.0 default, resolving both issues.

### DIFF
--- a/lib/knapsack_pro/runners/minitest_runner.rb
+++ b/lib/knapsack_pro/runners/minitest_runner.rb
@@ -14,6 +14,7 @@ module KnapsackPro
         end
 
         Rake::TestTask.new(task_name) do |t|
+          t.warning = false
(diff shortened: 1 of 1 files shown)


ArturT:
@patrickbizea I've released knapsack_pro 0.6.1 version. Thank you for PR!

ArturT:
I've released this fix also in knapsack gem https://github.com/ArturT/knapsack

PR #31, opened 2017-04-23
Add RSpec Queue Formatter to hide duplicated pending and failed tests in Queue Mode

ArturT:
* Add RSpec queue formatter and show summary of failed and pending tests at the end of queue mode run. 
* Hide pending and failed tests for intermediate tests runs after each fetched test files from Knapsack Pro API. In order to do that we monkey patch default rspec formatters.
* Allow user to disable monkey patching for default rspec formatters with flag `KNAPSACK_PRO_MODIFY_DEFAULT_RSPEC_FORMATTERS`

Related to: https://github.com/rspec/rspec-core/issues/2416

### DIFF
--- a/lib/knapsack_pro/formatters/rspec_queue_formatter.rb
+++ b/lib/knapsack_pro/formatters/rspec_queue_formatter.rb
@@ -1,0 +1,97 @@
+RSpec::Support.require_rspec_core('formatters/base_formatter')
+RSpec::Support.require_rspec_core('formatters/base_text_formatter')
+
+module KnapsackPro
+  module Formatters
+    module RSpecHideFailuresAndPendingExtension
+      def dump_failures(notification); end
+      def dump_pending(notification); end
+      def dump_summary(summary); end
+    end
+
+    class RSpecQueueFormatter < RSpec::Core::Formatters::BaseFormatter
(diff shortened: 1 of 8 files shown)


lib/knapsack_pro/formatters/rspec_queue_formatter.rb
```diff
@@ -0,0 +1,97 @@
...
+        return if notification.pending_examples.empty?
+        self.class.most_recent_pending = notification.fully_formatted_pending_examples
+      end
+    end
+  end
+end
+
+if KnapsackPro::Config::Env.modify_default_rspec_formatters?
+  class RSpec::Core::Formatters::BaseTextFormatter
+    prepend KnapsackPro::Formatters::RSpecHideFailuresAndPendingExtension
```
ArturT: http://stackoverflow.com/a/4471202/905697

toolmantim:
Thanks so much for this @ArturT! It's so much better.

One thing I'm still noticing is duplicate `--profile` output. Perhaps profile output should be disabled for queue mode?

ArturT:
@toolmantim I did fix for `--profile` and if you use profile formatter then you will see only single summary at the very end of the RSpec Queue Mode output.

Please update knapsack_pro to the newest version. ;)

toolmantim:
Amazing! ❤️

Issue #40, opened 2017-05-17
JUnit files are appended to with RSpec formatters


bobbytables:
Hello, found an interesting case while using this gem.

I'm trying to add JUnit formatter to use the results in Jenkins using the `junit` instruction. When using queue mode with this gem (and service) however, Knapsack might run rspec tests multiple times from the same process. This is causes a problem because `--out` in RSpec _appends_ to a file. So that means you end up with a file that has something along the lines of:

```
<xml ...>
<testcase stuff in here />
<xml> # next suite run starts and appends...
```

Jenkins (and from what I can see, XML parsers in general) don't like when there are multiple `<xml>` tags in the file, as you can see from this error:

```
Failed to read test report file /workspace/HCM_PR-7675-355EZNVL3NLOS25MQ7X2QIWUKL7VGXOWVAWXENJHPGTJ4IJEY4FA@5/tmp/rspec_results7c4385c84.xml
org.dom4j.DocumentException: Error on line 2175 of document file:///workspace/HCM_PR-7675-355EZNVL3NLOS25MQ7X2QIWUKL7VGXOWVAWXENJHPGTJ4IJEY4FA@5/tmp/rspec_results7c4385c84.xml : The processing instruction target matching "[xX][mM][lL]" is not allowed.
```

The more important piece is this:

```
The processing instruction target matching "[xX][mM][lL]" is not allowed.
```

Detail on the error is in this stackoverflow: http://stackoverflow.com/questions/19889132/error-the-processing-instruction-target-matching-xxmmll-is-not-allowed

I can see an "after hook" of sorts be helpful here because we could just rename the file afterwards to something that won't be appended to and Jenkins can read the multiple files just fine.

Thoughts?

ArturT:
Hey @bobbytables 

I could create custom JUnit formatter. I did similar thing for summary:
https://github.com/KnapsackPro/knapsack_pro-ruby/blob/master/lib/knapsack_pro/formatters/rspec_queue_summary_formatter.rb
in this PR:
https://github.com/KnapsackPro/knapsack_pro-ruby/pull/31/files

After that the formatter could be specified in the knapsack command similar as here:
https://github.com/KnapsackPro/knapsack_pro-ruby#how-to-use-junit-formatter

ArturT:
[...]

PR #41, opened 2017-05-22
Add after subset queue hook and example how to use JUnit formatter in Queue Mode

ArturT:
Related to issue: https://github.com/KnapsackPro/knapsack_pro-ruby/issues/40

Here is example of hook usage: https://github.com/KnapsackPro/rails-app-with-knapsack_pro/pull/6


lib/knapsack_pro/hooks/queue.rb
```diff
@@ -0,0 +1,21 @@
+module KnapsackPro
+  module Hooks
+    class Queue
+      class << self
+        attr_reader :after_subset_queue
+
+        def after_subset_queue(&block)
+          @after_subset_queue ||= block
+        end
```
bobbytables: This solution would at least allow us to rename the file that we output the results to. So I think this is actually a fine solution. Because in our spec helper we can do:

```
KnapsackPro::Hooks::Queue.after_subset_queue do |queue_id, subset_queue_id|
  FileUtils.mv('tmp/rspec.xml', "tmp/rspec_#{SecureRandom.hex(10)}.xml")
end

ArturT: Yep, I tried that here https://github.com/KnapsackPro/rails-app-with-knapsack_pro/pull/6/files but there is another problem described here https://github.com/KnapsackPro/knapsack_pro-ruby/issues/40#issuecomment-303519352

PR #55, opened 2017-11-08
Add support for Heroku CI environment variables

ArturT:
To test branch:

```
# Gemfile
gem 'knapsack_pro', github: 'KnapsackPro/knapsack_pro-ruby', branch: 'heroku-ci'
```

Heroku docs:
https://devcenter.heroku.com/articles/heroku-ci#quick-start
https://devcenter.heroku.com/articles/heroku-ci-parallel-test-runs


lib/knapsack_pro/config/ci/heroku.rb
```diff
@@ -0,0 +1,31 @@
...
+        def commit_hash
+          ENV['HEROKU_TEST_RUN_COMMIT_VERSION']
+        end
+
+        def branch
+          ENV['HEROKU_TEST_RUN_BRANCH']
+        end
+
+        def project_dir
+          '/app' if node_build_id
```
ArturT: I assume the project directory is always in `/app` directory on Heroku CI.

lib/knapsack_pro/config/ci/heroku.rb
```diff
@@ -0,0 +1,31 @@
+module KnapsackPro
+  module Config
+    module CI
+      class Heroku < Base
+        def node_total
+          # not provided
```
ArturT: It would be great to be able to read from Heroku ENV info how many parallel CI nodes we have. For instance we have 5 parallel jobs on Heroku then the value for `node_total` should be `5`.

lib/knapsack_pro/config/ci/heroku.rb
```diff
@@ -0,0 +1,31 @@
...
+        end
+
+        def node_index
+          # not provided
```
ArturT: Here we should be able to read from Heroku ENV the index of current parallel node. It should start from 0. So in case when we have 5 node total then the node_index is within range 0 - 4.

If Heroku won't expose this node index then we can manually set `KNAPSACK_PRO_CI_NODE_INDEX`

lib/knapsack_pro/config/ci/heroku.rb
mikescar: It is as far as I know, and `.git` is not present.

ArturT: @mikescar Ok. Actually, we don't need to care about `project_dir` when Heroku provides branch and git commit via ENV vars. I use `project_dir` only when I need to detect branch/git hash because they are not available via ENV for CI provider. In case of Heroku they expose branch/git hash so we should be good. The knapsack_pro won't use `project_dir`.

gudmundur:
Looks good to me. 👍

PR #62, opened 2018-06-23
Fix to record proper time for around(:each) in RSpec 

ArturT:
Use around :each so thanks to that we can track time for around :each defined by user in spec_helper or rails_helper


ArturT:
Related old PR: https://github.com/KnapsackPro/knapsack_pro-ruby/pull/52

Issue #59, opened 2018-03-13
Add test file exclude pattern


ArturT:
# Example:

When you would like to run tests only from spec/features directory then run:

```
KNAPSACK_PRO_TEST_SUITE_TOKEN_RSPEC=$API_TOKEN_FOR_FEATURE_TESTS \
KNAPSACK_PRO_TEST_DIR=spec \
KNAPSACK_PRO_TEST_FILE_PATTERN="spec/features/**{,/*/**}/*_spec.rb" \
bundle exec rake knapsack_pro:queue:rspec
```

# Problem:

What if you would like to run all tests except the `spec/features` directory?

We could use `KNAPSACK_PRO_TEST_FILE_PATTERN` and use a pattern that works with `Dir.glob` but sometimes it's hard to exclude something.

For now users have to specify all directories they want to run tests from. 
https://github.com/KnapsackPro/knapsack_pro-ruby#how-to-run-only-rspec-feature-tests-or-non-feature-tests

# Idea (todo):

We could add new variable `KNAPSACK_PRO_TEST_FILE_EXCLUDE_PATTERN`
and then you could do this:

```
files = Dir.glob(KNAPSACK_PRO_TEST_FILE_PATTERN)
files -= Dir.glob(KNAPSACK_PRO_TEST_FILE_EXCLUDE_PATTERN)
```

ArturT:
This issue is related: https://github.com/KnapsackPro/knapsack_pro-ruby/issues/57

ArturT:
I added exclude pattern in this PR https://github.com/KnapsackPro/knapsack_pro-ruby/pull/72 and I've just released a new version of knapsack_pro gem.

Here is documentation for exclude pattern:
https://github.com/KnapsackPro/knapsack_pro-ruby#how-to-exclude-tests-from-running-them

PR #72, opened 2018-10-28
Add test file exclude pattern

ArturT:
Related issue: https://github.com/KnapsackPro/knapsack_pro-ruby/issues/59


ArturT:
Here is an alternative solution prepared by one of knapsack_pro users:
https://gist.github.com/ddiachkov/d70780113174a2ff1fc65eaf9cb847a2

PR #87, opened 2019-06-07
Add Queue Mode for Cucumber

ArturT:
# Changes
This adds Queue Mode for Cucumber test runner so you can run tests with:

```bash
bundle exec rake knapsack_pro:queue:cucumber
```

Also it's possible to use junit formatter to get junit xml reports

```bash
bundle exec rake "knapsack_pro:queue:cucumber[--format junit --out tmp/test-reports/cucumber/queue_mode/]"
```

# Supported Queue Hooks in Cucumber Queue Mode

* supported: `KnapsackPro::Hooks::Queue.before_queue` and `KnapsackPro::Hooks::Queue.after_subset_queue`
* NOT supported: `KnapsackPro::Hooks::Queue.after_queue`

```ruby
# features/support/knapsack_pro.rb
KnapsackPro::Hooks::Queue.before_queue do |queue_id|
  print 'Before Queue Hook - run before test suite'
end

KnapsackPro::Hooks::Queue.after_subset_queue do |queue_id, subset_queue_id|
  print 'After Subset Queue Hook - run after subset of test suite'
end

# this hook is not supported and won't run
KnapsackPro::Hooks::Queue.after_queue do |queue_id|
  print 'After Queue Hook - run after test suite'
end
```

# Additional info for development of this gem
## why we don't use rake task
* We don't use Rake task invoke/execute because when test fails then the whole process exits immediately and because of that we can't run another set of tests from Queue.

```ruby
          task_name = "knapsack_pro:cucumber_run:#{ENV.fetch('KNAPSACK_PRO_SUBSET_QUEUE_ID')}"
          Cucumber::Rake::Task.new(task_name) do |t|
            t.cucumber_opts = "#{args} --require #{runner.test_dir} -- #{stringify_test_file_paths}"
          end
          #Rake::Task[task_name].invoke
          Rake::Task[task_name].execute
``` 

Instead we call `bundle exec cucumber` with arguments to run Cucumber for each set of tests fetched from Queue.


sergiopantoja:
This is awesome!

ArturT:
@sergiopantoja Thanks for checking it. Let's connect https://www.linkedin.com/in/arturtrzop/

Do you already have an account to test this cucumber integration?

sergiopantoja:
@ArturT Yup. We have the cucumber queue running in our CI already and it seems to be working great. Added you on LinkedIn. 👍

PR #102, opened 2020-03-14
Split RSpec test files by test examples

ArturT:
## Split test files by test cases

__Note:__ this is an experimental feature. It works for Regular Mode and Queue Mode. For large test suite with a few thousand test files, it may generate too many RSpec test example paths that may lead to too large JSON payload in request to Knapsack Pro API and this could trigger the API timeout.

Please give us feedback so we could improve the feature.
https://knapsackpro.com/contact

__How it works__: You can split slow test file by test cases. Thanks to that the test file can be split across parallel CI nodes because test cases from the test file will run on different CI nodes.

This is helpful when you have one or a few very slow test files that are a bottleneck for CI build speed and you don't want to manually create a few smaller test files from the slow test files. Instead, you can tell `knapsack_pro` gem to split your test files by test cases across parallel CI nodes.

### RSpec split test files by test examples (by individual `it`s)

In order to split RSpec test files by test examples across parallel CI nodes you need to set flag:

```
KNAPSACK_PRO_RSPEC_SPLIT_BY_TEST_EXAMPLES=true
```

Thanks to that your CI build speed can be faster. We recommend using this feature with Queue Mode to ensure parallel CI nodes finish work at a similar time which gives you the shortest CI build time.

Doing tests split by test examples can generate a lot of logs by `knapsack_pro` gem in Queue Mode. We recommend to set log level to:

```
KNAPSACK_PRO_LOG_LEVEL=warn
```

---
Story https://trello.com/c/qfUaxAg4 (gem maintainers only)


lib/knapsack_pro/base_allocator_builder.rb
```diff
@@ -38,7 +38,27 @@ def test_file_pattern
...
+
+        KnapsackPro.logger.warn("Generating RSpec test examples JSON report to prepare your test suite to be split by test examples (by individual 'it'. Thanks to that a single test file can be split across parallel CI nodes). Analyzing #{test_files_count} test files.")
+
+        if test_files_count > 1000
+          KnapsackPro.logger.warn("You have more than 1000 test files, it may take longer to generate test examples. Please wait...")
+        end
+
+        # generate RSpec JSON report in separate process to not pollute RSpec state
+        cmd = 'bundle exec rake knapsack_pro:rspec_test_example_detector'
+        Kernel.system(cmd)
```
shadre, line 54: I wonder if we should have a backup in case this fails. We could react here differently depending on the return value. (For reference, according to the docs: `true` means zero exit status, `false` - non-zero status and `nil` means that the execution failed).

lib/knapsack_pro/runners/queue/rspec_runner.rb
```diff
@@ -16,7 +16,9 @@ def self.run(args)
           cli_args = (args || '').split
           # if user didn't provide the format then use explicitly default progress formatter
           # in order to avoid KnapsackPro::Formatters::RSpecQueueSummaryFormatter being the only default formatter
-          cli_args += ['--format', 'progress'] unless cli_args.include?('--format')
+          if !cli_args.include?('--format') && !cli_args.include?('-f')
```
shadre, line 19: :+1:

lib/knapsack_pro/base_allocator_builder.rb
shadre: Sorry, I just saw we are raising an error in the other class in such a case, so scratch that.^

lib/knapsack_pro/base_allocator_builder.rb
```diff
@@ -38,7 +38,29 @@ def test_file_pattern
...
+
+        KnapsackPro.logger.warn("Generating RSpec test examples JSON report to prepare your test suite to be split by test examples (by individual 'it'. Thanks to that a single test file can be split across parallel CI nodes). Analyzing #{test_files_count} test files.")
+
+        if test_files_count > 1000
+          KnapsackPro.logger.warn("You have more than 1000 test files, it may take longer to generate test examples. Please wait...")
+        end
+
+        # generate RSpec JSON report in separate process to not pollute RSpec state
+        cmd = 'bundle exec rake knapsack_pro:rspec_test_example_detector'
+        unless Kernel.system(cmd)
```
ArturT, line 54: @shadre I added it anyway. `detector.test_file_example_paths` method can raise when json file missing.

lib/knapsack_pro/test_case_detectors/rspec_test_example_detector.rb
```diff
@@ -0,0 +1,68 @@
...
+
+      def test_file_example_paths
+        raise "No report found at #{REPORT_PATH}" unless File.exists?(REPORT_PATH)
+
+        json_report = File.read(REPORT_PATH)
+        hash_report = JSON.parse(json_report)
+        hash_report
+          .fetch('examples')
+          .map { |e| e.fetch('id') }
+          .map { |path| test_file_hash_for(path) }
```
shadre, line 36: What do you think about calling the block variable `path_with_example_id` just so that it's clear?

ArturT: I'm not sure what do you mean?

Something like that

```
.map(&:test_file_hash_for)
```

shadre: I just meant substituting `|path|` with `|path_with_example_id|` :)

PR #111, opened 2020-05-02
Raise error from Cucumber Queue Mode runner when system process doesn't finish execution

shadre:
# Problem

Sometimes Cucumber in Queue Mode could be killed by CI server for some reason (like exceeding the memory).
We want to make sure the user gets meaningful error as to what happened (and be double sure that killing the process never results in a false positive).

# Fix

We raise exception when Cucumber process in Queue Mode failed to execute properly (never exited on its own).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Change Log
 
+### 1.20.2
+
+* Raise an error when running Cucumber in Queue Mode and Cucumber system process doesn't finish execution correctly (for instance Cucumber process was killed by CI server due to lack of memory)
+
+    https://github.com/KnapsackPro/knapsack_pro-ruby/pull/111
+
+https://github.com/KnapsackPro/knapsack_pro-ruby/compare/v1.20.1...v1.20.2
+
 ### 1.20.1
 
 * Fix bug in RSpec split by test examples in < RSpec 3.6.0 (related to custom json formatter)
--- a/lib/knapsack_pro/runners/queue/cucumber_runner.rb
+++ b/lib/knapsack_pro/runners/queue/cucumber_runner.rb
@@ -91,7 +91,16 @@ module KnapsackPro
           # which is defined in lib/knapsack_pro/adapters/cucumber_adapter.rb
           ENV['KNAPSACK_PRO_BEFORE_QUEUE_HOOK_CALLED'] = 'true'
 
-          $?.exitstatus
+          process_status = $?
+
+          unless process_status.exited?
+            raise "Cucumber process execution failed. It's likely that your CI server has exceeded"\
+                    " its available memory. Please try changing CI config or retrying the CI build.\n"\
+                    "Failed command: #{cmd}\n"\
+                    "Process status: #{process_status.inspect}"
+          end
+
+          process_status.exitstatus
         end
       end
     end
--- a/spec/knapsack_pro/runners/queue/cucumber_runner_spec.rb
+++ b/spec/knapsack_pro/runners/queue/cucumber_runner_spec.rb
@@ -117,36 +117,49 @@ describe KnapsackPro::Runners::Queue::CucumberRunner do
 
         expect(ENV).to receive(:[]=).with('KNAPSACK_PRO_BEFORE_QUEUE_HOOK_CALLED', 'true')
 
-        expect($?).to receive(:exitstatus).and_return(exitstatus)
+        expect($?).to receive(:exited?).and_return(process_exited)
+        allow($?).to receive(:exitstatus).and_return(exitstatus)
       end
 
-      context 'when tests are passing' do
-        let(:exitstatus) { 0 }
+      context 'when system process finished its work (exited)' do
+        let(:process_exited) { true }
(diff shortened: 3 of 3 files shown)


lib/knapsack_pro/runners/queue/cucumber_runner.rb
```diff
@@ -91,7 +91,16 @@ def self.cucumber_run(runner, test_file_paths, args)
           # which is defined in lib/knapsack_pro/adapters/cucumber_adapter.rb
           ENV['KNAPSACK_PRO_BEFORE_QUEUE_HOOK_CALLED'] = 'true'
 
-          $?.exitstatus
+          process_status = $?
+
+          if !process_status.exited?
```
ArturT, line 96: if could be unless here

lib/knapsack_pro/runners/queue/cucumber_runner.rb
```diff
@@ -91,7 +91,16 @@ def self.cucumber_run(runner, test_file_paths, args)
...
+            raise "Cucumber process execution failed. It's likely that your CI server has exceeded"\
+                    " its available memory. Please try changing CI config or retrying the CI build.\n"\
+                    "Failed command: #{cmd}\n"\
+                    "Proces status: #{process_status.inspect}"
```
ArturT, line 100: typo: Process

spec/knapsack_pro/runners/queue/cucumber_runner_spec.rb
```diff
@@ -117,36 +117,50 @@
 
         expect(ENV).to receive(:[]=).with('KNAPSACK_PRO_BEFORE_QUEUE_HOOK_CALLED', 'true')
 
-        expect($?).to receive(:exitstatus).and_return(exitstatus)
+        expect($?).to receive(:exited?).and_return(process_exited)
+        allow($?).to receive(:exitstatus).and_return(exitstatus)
```
ArturT, line 121: I guess here we could have allow on `process_exited` to be more aligned with code where we call `process_status.exitstatus`

spec/knapsack_pro/runners/queue/cucumber_runner_spec.rb
```diff
@@ -117,36 +117,50 @@
...
-            can_initialize_queue: false,
-            args: args,
-            exitstatus: exitstatus,
-            all_test_file_paths: test_file_paths,
-          })
+      context 'when system process finished its work (exited)' do
+        let(:process_exited) { true }
+
+        context 'when tests are passing' do
+          let(:process_exited) { true }
```
ArturT, line 128: no need to duplicated here `let(:process_exited) { true }`. It's already defined in higher context.

ArturT:
We can add some context in the description of this PR like. This can be useful if someone will check this PR to understand why we did the change.

```
# Problem

example edge case when Cucumber can be killed by CI. Explain that this could lead to situation that tests was green but never run due to Cucumber process killed.

# Fix

We raise exception when Cucumber process was killed.
```

spec/knapsack_pro/runners/queue/cucumber_runner_spec.rb
shadre: yeah, forgot to remove, thanks.

spec/knapsack_pro/runners/queue/cucumber_runner_spec.rb
shadre: Do you mean changing this to `allow(process_status).to receive(:exited?)`? If so, then I'm not sure I agree with that as that would mean the test knows the internals of the tested class too much in my opinion.

By mocking `$?`'s method, we essentially just mock the outside world and don't care how it's implemented in the class. In other words, I don't believe this test should start failing when we change the naming of the local variable in the class.

ArturT: right, I wrongly assumed we already have mocked $? here but it's not needed. No changes here.

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Change Log
 
+### 1.20.2
+
+* Raise error from Cucumber Queue Mode runner when system process doesn't finish execution
+
+    https://github.com/KnapsackPro/knapsack_pro-ruby/pull/111
+
+https://github.com/KnapsackPro/knapsack_pro-ruby/compare/v1.20.1...v1.20.0
```
ArturT, line 9: there should be v1.20.2 at the end

shadre: missed that :( thanks for catching

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Change Log
 
+### 1.20.2
+
+* Raise error from Cucumber Queue Mode runner when system process doesn't finish execution
```
ArturT, line 5: Let's change this to a more explicit example:

Raise an error when running Cucumber in Queue Mode and Cucumber system process doesn't finish execution correctly (for instance Cucumber process was killed by CI server due to lack of memory)

PR #115, opened 2020-06-22
Support for non-delimited formatting params of rspec

scream3:
###  Bug
Passing formatting params to rspec without delimiter results in redundantly added `progress` formatter.

### Fix
Check for rspec params prefix instead of inclusion check.


ArturT:
Hi @scream3 , thanks for PR. Good catch! I will merge it and release a new version of knapsack_pro soon.

ArturT:
@scream3 I've released knapsack_pro 1.22.3 :)

PR #124, opened 2020-09-10
When you use Regular Mode then try 6 attempts to connect to the API instead of 3 attempts

ArturT:
# problem

Before this PR the knapsack_pro gem does 3 attempts to connect to Knapsack Pro API before starting running tests in [Fallback Mode](https://github.com/KnapsackPro/knapsack_pro-ruby#what-happens-when-knapsack-pro-api-is-not-availablenot-reachable-temporarily).

The described problem below is relevant only for running knapsack_pro in Regular Mode.

## Example scenarios

Assume:
* there are 2 parallel CI nodes. 
* the user uses knapsack_pro gem in Regular Mode.

Here you can learn about the [difference between Regular Mode and Queue Mode](https://docs.knapsackpro.com/2020/how-to-speed-up-ruby-and-javascript-tests-with-ci-parallelisation)

## Scenario when Knapsack Pro API is not available at all (no bug - successful scenario)

* CI node index 0 can't connect to the API so it starts running tests in Fallback Mode.
* CI node index 1 can't connect to the API so it starts running tests in Fallback Mode.

The whole test suite is executed across parallel CI nodes. Everything works fine.

## Scenario when Knapsack Pro API was not available only for one of the parallel CI nodes (bug exists - buggy scenario)

* CI node index 0 can't connect to the API so it starts running tests in Fallback Mode
* CI node index 1 can connect to the API so it starts running tests based on list of tests from the API

There is a risk that tests that supposed to be run on CI node index 0 and were never fetched from API won't run at all because Fallback Mode can run a different set of tests on CI node index 0. 
CI node index 1 instead of running tests in Fallback Mode was able to connect to the API so it got a different set of tests then it would get if it was also running Fallback Mode.

Problem: 

* A) This can lead to the scenario that not all test files from test suite will be run as part of CI build. This is a problem.
* B) Some of test files run in Fallback Mode on CI node index 0 can be the same as test files fetched from API on CI node index 1. This is not an issue that we run some tests at least once. The biggest issue is that we can skip some test files (issue A).

**Important:** knapsack_pro gem in Regular Mode can guarantee each test file from the test suite will be executed only when Fallback Mode starts on all parallel CI nodes.

# what we aim for

* We should aim to the scenario that all parallel CI nodes can connect to Knapsack Pro API.
* Or we should aim to the scenario that all parallel CI nodes should run Fallback Mode.

Only those 2 scenarios guarantee that all test files from the test suite will be run as part of the CI build.

* We should avoid scenario that some of CI nodes connect to Knapsack Pro API and other CI nodes use Fallback Mode. This can lead to skipping some tests as part of CI build.

# solution

We can increase max request retries attempts to the Knapsack Pro API. 

knapsack_pro gem in Regular Mode will do 6 attempts to connect to the API before starting Fallback Mode.
6 attempts will be spread across 2 minutes. This should be plenty of time for the Knapsack Pro API to auto-scale up the new available servers to serve increased traffic to the API.

If in 2 minutes knapsack_pro gem can't connect to the API then it is more likely a serious problem with the API availability so API should rather be not available at all to all parallel CI nodes (then all CI nodes should start in Fallback Mode). 

Of course, there is still a risk that edge case can happen and across all parallel CI nodes some of them connected to API and some of them did 6 requests attempts and started Fallback Mode. 

## recommendations

* If you would like to be 100% sure you never have this edge case then you can completely disable Fallback Mode with the env variable `KNAPSACK_PRO_FALLBACK_MODE_ENABLED=false`.

* If you want to define on your own how many request attempts can be made you can do it with the env variable  `KNAPSACK_PRO_MAX_REQUEST_RETRIES=7`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Change Log
 
+### 2.2.2
+
+* When you use Regular Mode then try 6 attempts to connect to the API instead of 3 attempts
+
+    https://github.com/KnapsackPro/knapsack_pro-ruby/pull/124
+
+https://github.com/KnapsackPro/knapsack_pro-ruby/compare/v2.2.1...v2.2.2
+
 ### 2.2.1
 
 * Improve detection of test file path in test-unit runner for test files with shared examples
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 * When you use Regular Mode then try 6 attempts to connect to the API instead of 3 attempts
 
+    Add `KNAPSACK_PRO_MAX_REQUEST_RETRIES` environment variable to let user define their own number of request retries to the API. It is useful to set it to `0` for [forked repos](https://knapsackpro.com/faq/question/how-to-make-knapsack_pro-works-for-forked-repositories-of-my-project) when you want to relay on Fallback Mode.
+
     https://github.com/KnapsackPro/knapsack_pro-ruby/pull/124
 
 https://github.com/KnapsackPro/knapsack_pro-ruby/compare/v2.2.1...v2.3.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,12 @@
 # Change Log
 
-### 2.2.2
+### 2.3.0
 
 * When you use Regular Mode then try 6 attempts to connect to the API instead of 3 attempts
 
     https://github.com/KnapsackPro/knapsack_pro-ruby/pull/124
 
-https://github.com/KnapsackPro/knapsack_pro-ruby/compare/v2.2.1...v2.2.2
+https://github.com/KnapsackPro/knapsack_pro-ruby/compare/v2.2.1...v2.3.0
 
 ### 2.2.1
 
--- a/README.md
+++ b/README.md
@@ -2320,11 +2320,19 @@ There are a few ways to reproduce tests executed on CI node in your development 
 
 ##### for knapsack_pro regular mode
 
-knapsack_pro gem will retry requests to Knapsack Pro API multiple times every few seconds til it switch to fallback behavior and it will split test files across CI nodes based on popular test directory names. When knapsack_pro starts fallback mode then you will see a warning in the output.
+knapsack_pro gem will retry requests to Knapsack Pro API multiple times every few seconds till it switchs to fallback behavior (Fallback Mode) and it will split test files across CI nodes based on popular test directory names. When knapsack_pro starts Fallback Mode then you will see a warning in the output.
+
+Note there is an unlikely scenario when some of the CI nodes may start in Fallback Mode but others don't and then it could happen that some of test files might be skipped. You should [read this to learn more](https://github.com/KnapsackPro/knapsack_pro-ruby/pull/124) and decide if you like to use Fallback Mode when running tests with knapsack_pro Regular Mode.
+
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
 # Change Log
 
+### 2.3.0
+
+* When you use Regular Mode then try 6 attempts to connect to the API instead of 3 attempts
+
+    Add `KNAPSACK_PRO_MAX_REQUEST_RETRIES` environment variable to let user define their own number of request retries to the API. It is useful to set it to `0` for [forked repos](https://knapsackpro.com/faq/question/how-to-make-knapsack_pro-works-for-forked-repositories-of-my-project) when you want to relay on Fallback Mode.
```
shadre, line 7: typo: `relay on` should be `rely on`

README.md
```diff
@@ -2320,11 +2320,19 @@ There are a few ways to reproduce tests executed on CI node in your development
 
 ##### for knapsack_pro regular mode
 
-knapsack_pro gem will retry requests to Knapsack Pro API multiple times every few seconds til it switch to fallback behavior and it will split test files across CI nodes based on popular test directory names. When knapsack_pro starts fallback mode then you will see a warning in the output.
+knapsack_pro gem will retry requests to Knapsack Pro API multiple times every few seconds till it switchs to fallback behavior (Fallback Mode) and it will split test files across CI nodes based on popular test directory names. When knapsack_pro starts Fallback Mode then you will see a warning in the output.
```
shadre, line 2323: typo: `switchs` -> `switches`

README.md
```diff
@@ -2320,11 +2320,19 @@ There are a few ways to reproduce tests executed on CI node in your development
...
+knapsack_pro gem will retry requests to Knapsack Pro API multiple times every few seconds till it switches to fallback behavior (Fallback Mode) and it will split test files across CI nodes based on popular test directory names.
+
+Note that if one of the CI nodes will lose connection to Knapsack Pro API but other not then you may see that some of the test files will be executed on multiple CI nodes. **Fallback Mode guarantees each of the test files is run at least once across CI nodes when you use knapsack_pro in Queue Mode.** Thanks to that we know if the whole test suite is green or not. When knapsack_pro starts Fallback Mode then you will see a warning in the output.
```
shadre, line 2333: I propose to rephrase the first sentence to this:

`Note that if one of the CI nodes loses connection to Knapsack Pro API but others don't, then some of the test files may be executed on multiple CI nodes.`

lib/knapsack_pro/client/connection.rb
```diff
@@ -156,6 +155,22 @@ def get
           http.get(uri, json_headers)
         end
       end
+
+      def max_request_retries
+        # when user defined max request retries
+        return KnapsackPro::Config::Env.max_request_retries if KnapsackPro::Config::Env.max_request_retries
```
shadre, line 161: Do you think we should limit this setting in any way? E.g. we could do:
```ruby
return [KnapsackPro::Config::Env.max_request_retries, MAX_RETRY ].min if KnapsackPro::Config::Env.max_request_retries
```

ArturT: This might not be needed because for every next retry the waiting window is increasing by 8s so they should rather not flod our API.

See example waiting time in spec
https://github.com/KnapsackPro/knapsack_pro-ruby/blob/4be8142aa3cce05fdf5a82c793f3d857f9e0e1ad/spec/knapsack_pro/client/connection_spec.rb#L175,L204

shadre: Oh, good to know, thank you!

PR #125, opened 2020-09-17
Update list of non encryptable branches

ArturT:
Add new branch names like main, development etc. There is no need to encrypt them. Thanks to that Knapsack Pro API can use them as default branches to get from them recorded tests timing data that could be used for a fresh new branch CI build.

Add branches with prefix refs/heads/ that are used on Github Actions.


shadre (commented):
:+1:

PR #136, opened 2020-12-14
Fix bug when failure_notifications is nil for RSpec formatter in Queue Mode (RSpecQueueSummaryFormatter)

ArturT:


ArturT:
We couldn't resolve the issue on the user side. It looked like RSpec error itself (not a knapsack_pro issue).

arich:
@ArturT would you mind reopening this and restoring the branch so I could try using it again?

ArturT:
@arich I've restored the branch. :) Let me know if you find out anything. You can also fork the repo and use your own copy and test different code changes if you like.

```
# Gemfile for CI testing purpose
# use your forked repo name and your branch name below
gem 'knapsack_pro', github: 'KnapsackPro/knapsack_pro-ruby', branch: 'master'
```

Or another way for development environment:

```
# Gemfile for local testing purpose
# provide path to copy of knapsack_pro repo on your disk
gem 'knapsack_pro', path: '../knapsack_pro-ruby'
```

How to run knapsack_pro in development: 
https://knapsackpro.com/faq/question/how-to-run-tests-for-particular-ci-node-in-your-development-environment 

You can modify the source code of knapsack_pro gem https://github.com/KnapsackPro/knapsack_pro-ruby with pry gem https://github.com/pry/pry 
Add the below line in any of the source files of forked knapsack_pro gem to debug it:
`require 'pry'; binding.pry`

arich:
Thanks @ArturT!
It might have been user error 😅 
I removed an old spec that had `allow_any_instance_of(StandardError)` stubbed out, which seems like an easy way to break things. We might be okay now, I replaced it with some normal instance stubbing. It's passing locally, I think it's passing using knapsack_pro:rspec, but failing using queue mode. Still doing some more testing there. 

Thanks for your help!

arich:
I think you can re-delete this branch. We were able to avoid the problem by not mocking out and placing expecations on the `message` and `backtrace` methods on error objects. Test is cleaner and simpler now anyway 😄 

Thank you!

ArturT:
@arich Great you solved it! It reminds me of a bug that I made by mistake a few years ago when working for a client. I defined the backtrace method in one of my ruby class and this overrides the default ruby backtrace method and caused a production infinity loop between our app and external API (we run out of API requests limit per month in a few hours). It was a sleepless night to find the root problem and make production stable again. xD

arich:
Haha! Great story. Won't do that again!

PR #138, opened 2021-03-03
Update README.md - update info about KNAPSACK_PRO_REPOSITORY_ADAPTER

ArturT:
Small improvements regarding KNAPSACK_PRO_REPOSITORY_ADAPTER flag in readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -461,8 +461,8 @@ You can also use git as repository adapter to determine branch and commit hash, 
 
 #### When you don't use git
 
-If your CI provider does not expose commit hash and branch name as environment variable then `knapsack_pro` gem does not know what is commit hash and branch name.
-You can manually set below values as environment variables:
+If your CI provider does not expose commit hash and branch name through environment variables, then `knapsack_pro` gem does not know these values.
+You can manually set the values of the current commit hash and branch name in the environment variables:
 
 * `KNAPSACK_PRO_COMMIT_HASH` - commit hash.
 * `KNAPSACK_PRO_BRANCH` - branch name.
--- a/README.md
+++ b/README.md
@@ -72,8 +72,9 @@ We keep this old FAQ in README to not break old links spread across the web. You
     - [Set API key token](#set-api-key-token)
     - [Set knapsack_pro command to execute tests](#set-knapsack_pro-command-to-execute-tests)
   - [Repository adapter (How to set up 3 of 3)](#repository-adapter-how-to-set-up-3-of-3)
-    - [When you NOT set global variable `KNAPSACK_PRO_REPOSITORY_ADAPTER` (default)](#when-you-not-set-global-variable-knapsack_pro_repository_adapter-default)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -442,7 +443,7 @@ __Tip 2:__ If you use one of unsupported CI providers ([here is list of supporte
 
 ### Repository adapter (How to set up 3 of 3)
 
-#### When you NOT set global variable `KNAPSACK_PRO_REPOSITORY_ADAPTER` (default)
+#### When you do NOT set global variable `KNAPSACK_PRO_REPOSITORY_ADAPTER` (default)
```
shadre, line 446: Do we mean here "When you should NOT" set this? If so, then I would say "should NOT" instead of "do NOT".

README.md
```diff
@@ -452,12 +453,20 @@ By default `KNAPSACK_PRO_REPOSITORY_ADAPTER` variable has no value so knapsack_p
...
-#### When should you set global variable `KNAPSACK_PRO_REPOSITORY_ADAPTER=git` (required when CI provider is not supported)
+#### When should you set global variable `KNAPSACK_PRO_REPOSITORY_ADAPTER=git` (when CI provider is not supported and you use git)
 
 `KNAPSACK_PRO_REPOSITORY_ADAPTER` - When it has the value `git`, your local version of git on CI server will be used to get the branch name and commit hash. You also need to set `KNAPSACK_PRO_PROJECT_DIR` with the project directory path.
 
 `KNAPSACK_PRO_PROJECT_DIR` - Path to the project on the CI node, for instance `/home/ubuntu/my-app-repository`. It should be the top-level directory of your repository.
 
+#### When you don't use git
+
+If your CI provider does not expose commit hash and branch name as environment variable then `knapsack_pro` gem does not know what is commit hash and branch name.
```
shadre, line 464: typo: it should either be "through an environment variable" (if we would expect one variable for those valyes), or "through environment variables" (if two).

I would change the second part to ", then `knapsack_pro` does not know these values" to avoid repetition.

README.md
```diff
@@ -452,12 +453,20 @@ By default `KNAPSACK_PRO_REPOSITORY_ADAPTER` variable has no value so knapsack_p
...
+You can manually set below values as environment variables:
```
shadre, line 465: It occurred to me that we are really not listing _the values_ one should set here. Just the variables in which the actual values should be stored. So maybe something like that would work better?:

`You can manually set the values of the current commit hash and branch name in the environment variables:`

Or maybe `You can manually set these values...`.

shadre (approved):
Thanks for taking care of this.

I suggested some small improvements for you to consider.

README.md
ArturT: No.

This is default, when you do not set the global variable.

shadre: I'm not sure if I understand.

What I'm struggling with is, as a reader, I don't know what exactly you mean by saying "when you do NOT set..."

ArturT: Maybe the header could be rephrased to:

By default `KNAPSACK_PRO_REPOSITORY_ADAPTER` environment variable is undefined

shadre: Okay, sounds good

ArturT: Fixed in https://github.com/KnapsackPro/knapsack_pro-ruby/commit/1ba4f4bf6aa7836391d477032ed256edc540a034

PR #148, opened 2021-07-08
Do not allow to use the RSpec tag option together with the RSpec split by test examples feature in knapsack_pro gem in Regular Mode

ArturT:
Why don't allow to use RSpec tag option with [RSpec split by test examples feature](https://knapsackpro.com/faq/question/how-to-split-slow-rspec-test-files-by-test-examples-by-individual-it)?
https://knapsackpro.com/faq/question/how-to-split-slow-rspec-test-files-by-test-examples-by-individual-it#warning-dont-use-rspec-tag-option

If you do it and you use at the same time `KNAPSACK_PRO_RSPEC_SPLIT_BY_TEST_EXAMPLES=true` then `--tag` option might not be respected by RSpec due to [a bug in the RSpec](https://github.com/rspec/rspec-core/issues/2522). RSpec has a higher priority to run test examples like `spec/a_spec.rb[1:1]` than respect the `--tag` option. This can result in running test cases that you didn't intend to run.

# solution

We raise an exception when the user uses RSpec tag option at the same time with RSpec split by test examples feature to warn him. 

# related

We already implemented such an exception for Queue Mode in https://github.com/KnapsackPro/knapsack_pro-ruby/pull/139

### DIFF
--- a/spec/knapsack_pro/runners/queue/rspec_runner_spec.rb
+++ b/spec/knapsack_pro/runners/queue/rspec_runner_spec.rb
@@ -133,26 +133,11 @@ describe KnapsackPro::Runners::Queue::RSpecRunner do
       context 'when RSpec split by test examples feature is enabled' do
         before do
           expect(KnapsackPro::Config::Env).to receive(:rspec_split_by_test_examples?).and_return(true)
+          expect(KnapsackPro::Adapters::RSpecAdapter).to receive(:ensure_no_tag_option_when_rspec_split_by_test_examples_enabled!).and_call_original
         end
 
-        context 'when tag option is provided as --tag' do
+        context 'when tag option is provided' do
           let(:args) { '--tag example-value' }
-
-          it do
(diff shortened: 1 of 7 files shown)


spec/knapsack_pro/runners/queue/rspec_runner_spec.rb
```diff
@@ -133,31 +133,16 @@
...
 
-        context 'when tag option is provided as --tag' do
+        context 'when tag option is provided' do
           let(:args) { '--tag example-value' }
 
           it do
             expect { subject }.to raise_error(/It is not allowed to use the RSpec tag option together with the RSpec split by test examples feature/)
           end
         end
-
```
ArturT, line 145: One general integration test is enough, so I removed below test cases as they are covered for  `ensure_no_tag_option_when_rspec_split_by_test_examples_enabled!` method spec now.

shadre (approved):
Looking good :+1:

PR #152, opened 2021-07-23
Track all test files assigned to a CI node in Regular Mode including pending test files in order to retry proper set of tests on the retried CI node

ArturT:
# bug

* When you use Regular Mode and RSpec and you run first CI build then Knapsack Pro API returns a list of test files assigned to the CI node. If the test files have a pending test file then RSpec won't run it. This causes not measuring time execution for the pending test file. Because of that list of test files that is saved to Knapsack Pro API does not contain the pending test file.
* Problem happens when you retry CI node and you expect the same set of test files will be assigned to the CI node. But there is missing pending spec file which was not recorded during the first CI build run.

# fix

Track all test files assigned to the CI node and save them to Knapsack Pro API after tests are executed. Thanks to that pending test file will have default 0 seconds time execution. When you retry CI node then the pending test file will be assigned to the retired CI node properly.

Improvement was made for Regular Mode for all test runners (RSpec, Cucumber, Minitest, ...etc)

# related old PR for a similar change in Queue Mode
https://github.com/KnapsackPro/knapsack_pro-ruby/pull/71

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Change Log
 
+### 2.18.2
+
+* Track all test files assigned to a CI node in Regular Mode including pending test files in order to retry proper set of tests on the retried CI node
+
+    https://github.com/KnapsackPro/knapsack_pro-ruby/pull/152
+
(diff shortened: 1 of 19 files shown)


shadre (approved):
Nicely done :+1:

PR #153, opened 2021-07-30
Use RSpec `example` block argument instead of the global `RSpec.current_example`. This allows to run tests with the `async-rspec` gem.

ArturT:
* Use RSpec `example` from RSpec `each` hook instead of global `RSpec.current_example`. It allows running tests with async-rspec gem.
* breaking change: drop support for RSpec 2.x. It was already accidentally dropped due to added `context` RSpec hooks.

# Related
This PR is based on changes from knapsack gem:
https://github.com/KnapsackPro/knapsack/pull/117

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Change Log
 
+### 3.0.0
+
+* __(breaking change)__ Remove support for RSpec 2.x. This change was already done by accident in [the pull request](https://github.com/KnapsackPro/knapsack_pro-ruby/pull/143) when we added RSpec `context` hook available only for RSpec 3.x.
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # Change Log
 
+### 3.0.0
```
shadre, line 3: Lingo suggestions:
- the RSpec \`context\` hook, which is available only since RSpec 3.x
- Use RSpec \`example\` block argument instead of the global \`RSpec.current_example\`. This allows to run tests with the \`async-rspec\` gem

shadre (approved):
added lingo suggestions

PR #155, opened 2021-08-06
Use `.knapsack_pro` directory for temporary files instead of the `tmp` directory in the user's project directory

ArturT:
# Problem

So far knapsack_pro gem generated temporary files in the `tmp/knapsack_pro` directory but people sometimes have RSpec hooks to clean up the`tmp` folder in their project. This leads to removing files generated by knapsack_pro gem.

# Solution

Use a different directory for temporary files. knapsack_pro gem will generate files in the `.knapsack_pro` directory in the user project folder.

`.knapsack_pro` folder will be ignored by git out of the box because we generate `.knapsack_pro/.gitignore` file that ignores all content in the `.knapsack_pro` folder.


lib/knapsack_pro/report.rb
```diff
@@ -5,7 +5,7 @@ def self.save
 
       if test_files.empty?
         KnapsackPro.logger.warn("No test files were executed on this CI node.")
-        KnapsackPro.logger.debug("When you use knapsack_pro regular mode then probably reason might be very narrowed tests list - you run only tests with specified tag and there are fewer test files with the tag than node total number.")
+        KnapsackPro.logger.debug("When you use knapsack_pro Regular Mode, the reason for no tests executing might be a very narrow tests list. Most likely, you run only tests with a specified tag, and the number of test files with the tag was lower than the number of parallel CI nodes.")
```
shadre, line 8: Nice improvements. :+1: 

the last bit could be simplified to ", and there were fewer test files with the tag than parallel CI nodes."

lib/knapsack_pro/report.rb
```diff
@@ -46,9 +45,9 @@ def self.save_node_queue_to_api
 
       if test_files.size > 0 && measured_test_files.size == 0
         KnapsackPro.logger.warn("#{test_files.size} test files were executed on this CI node but the recorded time was lost due to:")
-        KnapsackPro.logger.warn("1. Probably you have a code (i.e. RSpec hooks) that clears tmp directory in your project. Please ensure you do not remove the content of tmp/knapsack_pro/queue/ directory between tests run.")
-        KnapsackPro.logger.warn("2. Another reason might be that you forgot to add Knapsack::Adapters::RSpecAdapter.bind in your rails_helper.rb or spec_helper.rb. Please follow the installation guide again: https://docs.knapsackpro.com/integration/")
-        KnapsackPro.logger.warn("3. All your tests are empty test files, are pending tests or have syntax error and could not be executed hence no measured time execution by knapsack_pro.")
+        KnapsackPro.logger.warn("1. Please ensure you do not remove the content of .knapsack_pro/queue/ directory between tests run.")
```
shadre, line 48: "the contents"

I would add _the_ before `.knapsack_pro/queue/`

Do you think it's a good idea to mention the more general, main `.knapsack_pro` directory here instead of the subdir?

lib/knapsack_pro/report.rb
```diff
@@ -46,9 +45,9 @@ def self.save_node_queue_to_api
...
+        KnapsackPro.logger.warn("2. Ensure you added Knapsack::Adapters::RSpecAdapter.bind in your rails_helper.rb or spec_helper.rb. Please follow the installation guide again: https://docs.knapsackpro.com/integration/")
```
shadre, line 49: I think this should read "you've added"

lib/knapsack_pro/report.rb
```diff
@@ -46,9 +45,9 @@ def self.save_node_queue_to_api
...
+        KnapsackPro.logger.warn("3. Another reason for this warning is that all your tests are empty test files, pending tests, or they have syntax errors, and the time execution was not recorded for them.")
```
shadre, line 50: I would say "Another potential reason" to emphasize this being a hypothesis and not a statement

lib/knapsack_pro/tracker.rb
```diff
@@ -110,6 +110,7 @@ def save_prerun_tests_report(hash)
     end
 
     def read_prerun_tests_report
+      raise "Report #{prerun_tests_report_path} doest not exist on the disk. Most likely, it was removed accidentally. Please report the bug to the support team." unless File.exists?(prerun_tests_report_path)
```
shadre, line 113: Maybe we can add a link here, or at least specify that we mean the Knapsack support team?

spec/knapsack_pro/config/temp_files_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+    let(:gitignore_file_path) { '.knapsack_pro/.gitignore' }
+
+    subject { described_class.temp_directory_path }
+
+    it 'returns temporary directory path' do
+      expect(subject).to eq '.knapsack_pro'
+    end
+
+    it 'creates .gitignore file' do
+      expect(File.exists?(gitignore_file_path)).to be false
```
shadre, line 12: FYI there is a nice DSL for expectations like that https://relishapp.com/rspec/rspec-expectations/v/2-0/docs/matchers/expect-change

e.g. `expect{Counter.increment}.to change{Counter.count}.from(0).to(1)`

(just wanted to share this in case you forgot about it, the current one is cool by all means)

spec/knapsack_pro/config/temp_files_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+      subject
+      expect(File.exists?(gitignore_file_path)).to be true
+    end
+
+    it '.gitignore file has content' do
```
shadre, line 17: Do I understand correctly that the most important line in the file's content is the one with the asterisk (`*`)? If so, then maybe it'd be better to check the inclusion of it instead?

spec/knapsack_pro/slow_test_file_determiner_spec.rb
```diff
@@ -67,7 +67,7 @@
 
     context 'when json report does not exist' do
       it do
-        expect { subject }.to raise_error(RuntimeError, 'Report with slow test files was not generated yet. If you have enabled split by test cases https://github.com/KnapsackPro/knapsack_pro-ruby#split-test-files-by-test-cases and you see this error it means that your tests accidentally cleaned up tmp/knapsack_pro directory. Please do not remove this directory during tests runtime!')
+        expect { subject }.to raise_error(RuntimeError, 'Report with slow test files was not generated yet. If you have enabled split by test cases https://github.com/KnapsackPro/knapsack_pro-ruby#split-test-files-by-test-cases and you see this error it means that your tests accidentally cleaned up .knapsack_pro directory. Please do not remove this directory during tests runtime!')
```
shadre, line 70: Suggestions for the error msg:

"_The_ report with slow test files"

I would surround the url with the parens so that it doesn't disrupt the sentence.

"_the_ .knapsack_pro directory"

lib/knapsack_pro/config/temp_files.rb
```diff
@@ -0,0 +1,46 @@
+module KnapsackPro
+  module Config
+    class TempFiles
+      def self.temp_directory_path
```
shadre, line 4: Have you considered separating the ensuring the temp directory exists from returning the path? Or maybe memoizing the result or sth?

Do you think there is a way to do this check once at some point, or would it always be succeptible to errors?

The reason I'm wondering about this is that it seems that we are using this `.temp_directory_path` method in many places to just get the correct path, but each time we do this it would now trigger an I/O read operation. Would be great to avoid that if it's avoidable.

shadre (commented):
Adding some questions and comments

lib/knapsack_pro/config/temp_files.rb
```diff
@@ -0,0 +1,46 @@
...
+        temp_files = new
+        temp_files.ensure_temp_directory_exists
+        temp_files.temp_directory_path
+      end
+
+      def ensure_temp_directory_exists
```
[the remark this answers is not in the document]
ArturT: When I move code to `private` section:

```ruby
private

      def ensure_temp_directory_exists
        unless File.exists?(gitignore_file_path)
          create_temp_directory
          create_gitignore_file
        end
      end

      # relative to the directory where you run knapsack_pro gem (user's project)
      def temp_directory_path
        '.knapsack_pro'
      end
```

then I get error:

```
$ rspec ./spec/knapsack_pro/config/temp_files_spec.rb

Randomized with seed 3043

KnapsackPro::Config::TempFiles
  .temp_directory_path
    .gitignore file has content (FAILED - 1)
    creates .gitignore file (FAILED - 2)
    returns temporary directory path (FAILED - 3)

Failures:

  1) KnapsackPro::Config::TempFiles.temp_directory_path .gitignore file has content
     Failure/Error: temp_files.ensure_temp_directory_exists

     NoMethodError:
       private method `ensure_temp_directory_exists' called for #<KnapsackPro::Config::TempFiles:0x00007f911c9ed6d0>
     # ./lib/knapsack_pro/config/temp_files.rb:8:in `temp_directory_path'
     # ./spec/knapsack_pro/config/temp_files_spec.rb:5:in `block (3 levels) in <top (required)>'
     # ./spec/knapsack_pro/config/temp_files_spec.rb:18:in `block (3 levels) in <top (required)>'
     # /Users/artur/.rvm/gems/ruby-3.0.2/gems/webmock-3.13.0/lib/webmock/rspec.rb:37:in `block (2 levels) in <top (required)>'
```

---

When I move code to `protected` I get error as well:

```
  1) KnapsackPro::Config::TempFiles.temp_directory_path creates .gitignore file
     Failure/Error: temp_files.ensure_temp_directory_exists

     NoMethodError:
       protected method `ensure_temp_directory_exists' called for #<KnapsackPro::Config::TempFiles:0x00007fbb43a47c88>
     # ./lib/knapsack_pro/config/temp_files.rb:8:in `temp_directory_path'
     # ./spec/knapsack_pro/config/temp_files_spec.rb:5:in `block (3 levels) in <top (required)>'
     # ./spec/knapsack_pro/config/temp_files_spec.rb:13:in `block (3 levels) in <top (required)>'
     # /Users/artur/.rvm/gems/ruby-3.0.2/gems/webmock-3.13.0/lib/webmock/rspec.rb:37:in `block (2 levels) in <top (required)>'
```

We call the following methods in class method `temp_directory_path` so they probably must be public.

```ruby
 temp_files.ensure_temp_directory_exists
 temp_files.temp_directory_path
```

lib/knapsack_pro/config/temp_files.rb
```diff
@@ -0,0 +1,46 @@
...
+      end
+
+      private
+
+      def create_temp_directory
+        FileUtils.mkdir_p(temp_directory_path)
+      end
+
+      def gitignore_file_path
+        File.join(temp_directory_path, '.gitignore')
```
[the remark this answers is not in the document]
ArturT: > Do I understand correctly, that we'll create our own .gitignore file in our new temp directory?

Yes.

> Does it mean that there would now need to be the .knapsack_pro directory commited in the project (with just the .gitignore file committed)?

No. Because we also ignore `.gitignore` file with `*` char in `.gitignore` file.

`.gitignore` file is created during gem runtime. It does not have to be committed into repo even when you run knapsack pro locally in development because all content of `.knapsack_pro` is ignored so it means for git the `.knapsack_pro` directory is empty. It's not possible to commit an empty directory in git so this way `.knapsack_pro` with the `.gitignore` file won't be committed.

lib/knapsack_pro/config/temp_files.rb
```diff
@@ -0,0 +1,46 @@
...
+  module Config
+    class TempFiles
+      def self.temp_directory_path
+        temp_files = new
+        temp_files.ensure_temp_directory_exists
+        temp_files.temp_directory_path
+      end
+
+      def ensure_temp_directory_exists
+        unless File.exists?(gitignore_file_path)
```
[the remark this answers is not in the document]
ArturT: Good catch. I've used `File.exist?` in the whole codebase.

lib/knapsack_pro/config/temp_files.rb
ArturT: I was thinking about it and I decided that checking if file exists on the disk is good enough. We already do some IO operations in queue mode like saving reports and it shouldn't have significant impact on performance.

If I add memorization to class method then this causes random tests failures depending on the order of tests executed because class method memorization is preserved for each call of class method.

For instance I could do:

```ruby
      def self.temp_directory_path
        @temp_directory_path ||=
          begin
            temp_files = new
            temp_files.ensure_temp_directory_exists
            temp_files.temp_directory_path
          end
      end
```

and then when I run tests I get randomly different tests failures:

```
[14:47:44] artur@Artur-MBP: ~/Documents/github/knapsack-pro/knapsack_pro-ruby temporary-files-directory! ruby-3.0.2
$ rspec ./spec/knapsack_pro/config/temp_files_spec.rb

Randomized with seed 42508

KnapsackPro::Config::TempFiles
  .temp_directory_path
    .gitignore file has content
    returns temporary directory path
    creates .gitignore file (FAILED - 1)

Failures:

  1) KnapsackPro::Config::TempFiles.temp_directory_path creates .gitignore file
     Failure/Error: expect(File.exist?(gitignore_file_path)).to be true

       expected true
            got false
     # ./spec/knapsack_pro/config/temp_files_spec.rb:14:in `block (3 levels) in <top (required)>'
     # /Users/artur/.rvm/gems/ruby-3.0.2/gems/webmock-3.13.0/lib/webmock/rspec.rb:37:in `block (2 levels) in <top (required)>'

Finished in 0.02793 seconds (files took 0.53206 seconds to load)
3 examples, 1 failure

Failed examples:

rspec ./spec/knapsack_pro/config/temp_files_spec.rb:11 # KnapsackPro::Config::TempFiles.temp_directory_path creates .gitignore file

Randomized with seed 42508


[14:47:46] artur@Artur-MBP: ~/Documents/github/knapsack-pro/knapsack_pro-ruby temporary-files-directory! ruby-3.0.2
$ rspec ./spec/knapsack_pro/config/temp_files_spec.rb

Randomized with seed 56734

KnapsackPro::Config::TempFiles
  .temp_directory_path
    returns temporary directory path
    .gitignore file has content (FAILED - 1)
    creates .gitignore file (FAILED - 2)

Failures:

  1) KnapsackPro::Config::TempFiles.temp_directory_path .gitignore file has content
     Failure/Error: expect(File.read(gitignore_file_path)).to include '# This directory is used by knapsack_pro gem for storing temporary files during tests runtime'

     Errno::ENOENT:
       No such file or directory @ rb_sysopen - .knapsack_pro/.gitignore
     # ./spec/knapsack_pro/config/temp_files_spec.rb:19:in `read'
     # ./spec/knapsack_pro/config/temp_files_spec.rb:19:in `block (3 levels) in <top (required)>'
     # /Users/artur/.rvm/gems/ruby-3.0.2/gems/webmock-3.13.0/lib/webmock/rspec.rb:37:in `block (2 levels) in <top (required)>'

  2) KnapsackPro::Config::TempFiles.temp_directory_path creates .gitignore file
     Failure/Error: expect(File.exist?(gitignore_file_path)).to be true

       expected true
            got false
     # ./spec/knapsack_pro/config/temp_files_spec.rb:14:in `block (3 levels) in <top (required)>'
     # /Users/artur/.rvm/gems/ruby-3.0.2/gems/webmock-3.13.0/lib/webmock/rspec.rb:37:in `block (2 levels) in <top (required)>'

Finished in 0.01396 seconds (files took 0.39825 seconds to load)
3 examples, 2 failures

Failed examples:

rspec ./spec/knapsack_pro/config/temp_files_spec.rb:17 # KnapsackPro::Config::TempFiles.temp_directory_path .gitignore file has content
rspec ./spec/knapsack_pro/config/temp_files_spec.rb:11 # KnapsackPro::Config::TempFiles.temp_directory_path creates .gitignore file

Randomized with seed 56734
```

I could disable memorization with custom argument `def self.temp_directory_path(cached: true)`. For the test environment I could use`cache: false` but I'm not sure if this is good idea to change method behaviour only to make tests pass :/

lib/knapsack_pro/config/temp_files.rb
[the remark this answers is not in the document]
shadre: Yes, this is because we are trying to access the private method from outside the object.

If we are adamant about the public method remaining a class method, we could change all the other methods to be [private class methods](https://medium.com/rubycademy/3-ways-to-make-class-methods-private-in-ruby-64b970e54613).

lib/knapsack_pro/config/temp_files.rb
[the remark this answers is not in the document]
shadre: Cool, I hoped it works the way you described it. Great to know it does indeed! :green_heart:

spec/knapsack_pro/config/temp_files_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+      expect(File.exist?(gitignore_file_path)).to be true
+    end
+
+    it '.gitignore file has content' do
+      subject
+
+      expect(File.read(gitignore_file_path)).to include '# This directory is used by knapsack_pro gem for storing temporary files during tests runtime'
+      expect(File.read(gitignore_file_path)).to include '# Ignore all files, and do not commit this directory into your repository.'
+      expect(File.read(gitignore_file_path)).to include '# Learn more at https://knapsackpro.com'
+      expect(File.read(gitignore_file_path)).to include '*'
```
shadre, line 27: Can we test it in a way that will ensure this is a whole line containing just the asterisk? Currently as I understand it, any asterisk anywhere in the file would make this expectation pass. Business-wise we need more than that.

And if we are testing for the line with the asterisk, then maybe we could improve the example description to reflect that (maybe just updating it so that it mentions "_correct_ content" or sth like that is good enough?)

BTW, I'd be fine with removing all of the other expectations (checking for the commented out content), as I don't think they add much.

spec/knapsack_pro/slow_test_file_determiner_spec.rb
```diff
@@ -67,7 +67,7 @@
 
     context 'when json report does not exist' do
       it do
-        expect { subject }.to raise_error(RuntimeError, 'Report with slow test files was not generated yet. If you have enabled split by test cases https://github.com/KnapsackPro/knapsack_pro-ruby#split-test-files-by-test-cases and you see this error it means that your tests accidentally cleaned up tmp/knapsack_pro directory. Please do not remove this directory during tests runtime!')
+        expect { subject }.to raise_error(RuntimeError, 'Report with slow test files was not generated yet. If you have enabled split by test cases https://github.com/KnapsackPro/knapsack_pro-ruby#split-test-files-by-test-cases and you see this error it means that your tests accidentally cleaned up the .knapsack_pro directory. Please do not remove this directory during tests runtime!')
```
shadre, line 70: gramatically, this should read "has not been generated yet"

I would also say either "_The_ report" or "_A_ report" depending on what makes more sense.

lib/knapsack_pro/config/temp_files.rb
```diff
@@ -0,0 +1,38 @@
...
+      def self.ensure_temp_directory_exists!
+        unless File.exist?(gitignore_file_path)
+          create_temp_directory
+          create_gitignore_file
+        end
+      end
+
+      private
+
+      def self.create_temp_directory
```
shadre, line 16: I would personally add a bang to this name and the one on L31 as well.

shadre (approved):
Well done :+1: 

Added a few comments about minor things.

PR #162, opened 2021-09-16
[wip] Add puts to print exit code and see when callbacks are called in Minitest

ArturT:
This is for debugging purpose.

### DIFF
--- a/lib/knapsack_pro/runners/queue/minitest_runner.rb
+++ b/lib/knapsack_pro/runners/queue/minitest_runner.rb
@@ -32,7 +32,15 @@ module KnapsackPro
             accumulator = run_tests(accumulator)
           end
 
+          puts '+'*100
+          puts "Ensure we return exit code: #{accumulator[:exitstatus]}"
+          puts '+'*100
+
           Kernel.exit(accumulator[:exitstatus])
+
+          puts '+'*100
+          puts "Kernel.exit called"
+          puts '+'*100
         end
 
         def self.run_tests(accumulator)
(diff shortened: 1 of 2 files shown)


lib/knapsack_pro/runners/queue/minitest_runner.rb
```diff
@@ -32,7 +32,15 @@ def self.run(args)
...
           end
 
+          puts '+'*100
+          puts "Ensure we return exit code: #{accumulator[:exitstatus]}"
+          puts '+'*100
+
           Kernel.exit(accumulator[:exitstatus])
+
+          puts '+'*100
+          puts "Kernel.exit called"
```
ArturT, line 42: This should never be printed because when we do `Kernal.exit` then the process will be closed.

ArturT:
There is a better way to debug why CI node freezed. See the script here:
https://docs.knapsackpro.com/ruby/troubleshooting/#knapsack-pro-hangs

Add the following script to `spec_helper.rb`. Thanks to that, you will see the backtrace showing what code line causes the hanging problem. Backtrace is visible only if the Ruby process got kill with `USR1` signal.

```ruby
# source https://github.com/rspec/rspec-rails/issues/1353#issuecomment-93173691
puts "Process pid: #{Process.pid}"

trap 'USR1' do
  threads = Thread.list

  puts
  puts "=" * 80
  puts "Received USR1 signal; printing all #{threads.count} thread backtraces."

  threads.each do |thr|
    description = thr == Thread.main ? "Main thread" : thr.inspect
    puts
    puts "#{description} backtrace: "
    puts thr.backtrace.join("\n")
  end

  puts "=" * 80
end
```

PR #165, opened 2021-12-22
add error message to RspecAdapter#bind

nbr:
Due to the way the adapter name is inferred in `KnapsackPro::Adapters::BaseAdapter`, `RspecAdapter#bind` will not write the bind file to the correct location and the command will fail. 

This adds an error message when `RspecAdapter#bind` is called, rather than silently inheriting the method from `RSpecAdapter`.


ArturT:
@nbr Thank you for your contribution. I've released a new knapsack_pro gem version 3.2.0.

PR #167, opened 2022-03-03
Raise exception when using `:focus` tag to avoid skipping RSpec tests

ArturT:
Raise exception only when RSpec inclusion rule includes `:focus` tag.

# problem description

When tests can be skipped by RSpec?

The problem occurs when at least one of the test files has `:focus` tag. And when you use 

```ruby
RSpec.configure do |c|
  c.filter_run_when_matching :focus
end
```

For instance, if a test file has content like this:

```ruby
describe ArticlesController, :focus do
  it '1' do
    expect(true).to be true
  end

  # another way to make test focused is to add prefix f to it (fit)
  fit '2' do
    expect(true).to be true
  end
end
```

Then this sets RSpec filter to `focus: true`. You can see in RSpec output this: `Run options: include {:focus=>true}`
and as a result we can see in the RSpec output: `All examples were filtered out` so tests without the tag are skipped.

All further test files fetched from Queue are going to be skipped due to this filter `:focus => true`.

# solution

Using the `:focus` tag should be a temporary change in the development environment when you only want to run tests with the `:focus` tag. In the CI environment, it is not recommended to use `:focus` tag because this can lead to skipping test cases. Especially it can lead to skipping random test file paths without `:focus` tag in the Knapsack Pro Queue Mode when the test files are loaded after the test file with `:focus` tag. Please remove `:focus` tags from your test files and commit the change and push to the repository. RSpec provides aliases `fit`, `fdescribe` and `fcontext` as a shorthand for `it`, `describe` and `context` with `:focus` metadata, making it easy to temporarily focus an example or group by prefixing an `f`. Please remove all occurrences of it or use an alternative solution: https://knapsackpro.com/faq/question/why-is-rspec-skipping-some-of-the-tests#if-you-have-too-many-test-files-with-focus-tag-and-you-dont-want-to-remove-the-tag-manually

### DIFF
--- a/lib/knapsack_pro/adapters/rspec_adapter.rb
+++ b/lib/knapsack_pro/adapters/rspec_adapter.rb
@@ -63,7 +63,7 @@ module KnapsackPro
               end
 
             if example.metadata[:focus] && KnapsackPro::Adapters::RSpecAdapter.rspec_configuration.filter.rules[:focus]
-              raise "We detected a test file path #{current_test_path} with a test using metadata `:focus` tag. RSpec might not run some tests in Queue Mode (causing random tests skipping problem). Please remove `:focus` tag from your codebase. See more: https://knapsackpro.com/faq/question/rspec-is-not-running-some-tests"
+              raise "We detected a test file path #{current_test_path} with a test using the metadata `:focus` tag. RSpec might not run some tests in the Queue Mode (causing random tests skipping problem). Please remove the `:focus` tag from your codebase. See more: https://knapsackpro.com/faq/question/rspec-is-not-running-some-tests"
             end
 
             example.run
--- a/lib/knapsack_pro/adapters/rspec_adapter.rb
+++ b/lib/knapsack_pro/adapters/rspec_adapter.rb
@@ -63,7 +63,7 @@ module KnapsackPro
               end
 
             if example.metadata[:focus] && KnapsackPro::Adapters::RSpecAdapter.rspec_configuration.filter.rules[:focus]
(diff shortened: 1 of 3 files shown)


lib/knapsack_pro/adapters/rspec_adapter.rb
```diff
@@ -62,6 +62,10 @@ def bind_time_tracker
...
+              raise "We detected a test file path #{current_test_path} with a test using metadata `:focus` tag. Using the `:focus` tag should be a temporary change in the development environment when you only want to run tests with the `:focus` tag. In the CI environment, it is not recommended to use `:focus` tag because this can lead to skipping test cases. Especially it can lead to skipping random test file paths without `:focus` tag in the Knapsack Pro Queue Mode when the test files are loaded after the test file with `:focus` tag. Please remove `:focus` tags from your test files and commit the change and push to the repository. RSpec provides aliases `fit`, `fdescribe` and `fcontext` as a shorthand for `it`, `describe` and `context` with `:focus` metadata, making it easy to temporarily focus an example or group by prefixing an `f`. Please remove all occurrences of it or use an alternative solution: https://knapsackpro.com/faq/question/why-is-rspec-skipping-some-of-the-tests#if-you-have-too-many-test-files-with-focus-tag-and-you-dont-want-to-remove-the-tag-manually"
```
shadre, line 66: I wonder if this needs to be so wordy, especially since we are providing the URL for more details. What do you think?

Also, if we agree on the new FAQ entry's contents we need to update the URL here.

ArturT: I was wondering about that as well.

Keeping actionable information here might be useful if the link to the FAQ becomes obsolete. For instance, someone won't upgrade gem for a long time and the link could become outdated if we move the FAQ to a different location.

> Maybe this could be described in 1-2 sentences instead?

Maybe this would work better:

```
We detected a test file path #{current_test_path} with a test using metadata `:focus` tag. RSpec might not run some tests in Queue Mode (causing random tests skipping problem).

Please remove `:focus` tag from your codebase. See more: https://knapsackpro.com/faq/question/rspec-is-not-running-some-tests
```

lib/knapsack_pro/adapters/rspec_adapter.rb
```diff
@@ -62,6 +62,10 @@ def bind_time_tracker
                 current_test_path
               end
 
+            if example.metadata[:focus] && KnapsackPro::Adapters::RSpecAdapter.rspec_configuration.filter.rules[:focus]
+              raise "We detected a test file path #{current_test_path} with a test using metadata `:focus` tag. RSpec might not run some tests in Queue Mode (causing random tests skipping problem). Please remove `:focus` tag from your codebase. See more: https://knapsackpro.com/faq/question/rspec-is-not-running-some-tests"
```
shadre, line 66: I think we are missing some _the_'s:
- using _the_ metadata :focus tag
- in _the_ Queue Mode
- Please remove _the_ :focus tag

PR #178, opened 2022-09-27
Improve the RSpec Queue Mode runner log output (add seed)

shadre:


lib/knapsack_pro/runners/queue/rspec_runner.rb
```diff
@@ -107,10 +109,14 @@ def self.run_tests(accumulator)
...
+        def self.adapter_class
+          KnapsackPro::Adapters::RSpecAdapter
+        end
+
         def self.log_rspec_command(cli_args, test_file_paths, type)
           case type
           when :subset_queue
-            KnapsackPro.logger.info("To retry the last batch of tests fetched from the API Queue, please run the following command on your machine. (If you use the `-- order random` option, remember to add correct `--seed 123` that you can find at the end of the RSpec output.)")
+            KnapsackPro.logger.info("To retry the last batch of tests fetched from the API Queue, please run the following command on your machine:")
           when :end_of_queue
```
shadre, line 120: This case (invoked on L63) remains unchanged. I didn't like any of the options I came up to make it work. I figured it's better to improve other logs for now, and alter this case if I think of a good-enough solution for it.

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Change Log
 
+### 3.4.1
+* Improve the RSpec Queue Mode runner log output (add seed)
```
ArturT, line 4: Please add a new line above.

ArturT: Please add below a link to the PR https://github.com/KnapsackPro/knapsack_pro-ruby/pull/178

This way, users can quickly access more relevant details about the release. Typically we have some context in the PR description with more details about what changed.

lib/knapsack_pro/runners/queue/rspec_runner.rb
```diff
@@ -151,6 +157,22 @@ def self.rspec_clear_examples
             ::RSpec.configuration.reset_filters
           end
         end
+
+        def self.args_with_seed_option_added_when_viable(args, rspec_runner)
+          order_option = adapter_class.order_option(args)
+
+          if order_option
+            # Don't add the seed option for order other than random, e.g. `defined`
+            return args unless order_option include?('rand')
```
ArturT, line 166: 1. should there be a dot?
`order_option.include?('rand')`

Are you calling the `include?`  method on the `order_option` variable, right?

2. I'm not sure, is this working with both options? `rand` and `random`?

From RSpec docs `$ rspec --help`:

```
  --order TYPE[:SEED]            Run examples by the specified order type.
                                         [defined]           examples and groups are run in the order they are defined
                                         [rand]              randomize the order of groups and examples
                                         [random]            alias for rand
                                         [random:SEED]       e.g. --order random:123
                                         [recently-modified] run the most recently modified files first
```

ArturT: Regarding 2. I see you already tested that in the `spec/knapsack_pro/adapters/rspec_adapter_spec.rb` file. Please ignore this question.

ArturT (commented):
I checked the code and left some minor comments.

I'm going to do manual testing and then I will do the final code review.

lib/knapsack_pro/runners/queue/rspec_runner.rb
shadre: > 1. should there be a dot?

Obviously, yes. Thanks for catching this.

lib/knapsack_pro/runners/queue/rspec_runner.rb
```diff
@@ -107,10 +109,14 @@ def self.run_tests(accumulator)
...
+        end
+
         def self.log_rspec_command(cli_args, test_file_paths, type)
           case type
           when :subset_queue
-            KnapsackPro.logger.info("To retry the last batch of tests fetched from the API Queue, please run the following command on your machine. (If you use the `-- order random` option, remember to add correct `--seed 123` that you can find at the end of the RSpec output.)")
+            KnapsackPro.logger.info("To retry the last batch of tests fetched from the API Queue, please run the following command on your machine:")
           when :end_of_queue
             KnapsackPro.logger.info("To retry all the tests assigned to this CI node, please run the following command on your machine:")
           end
```
ArturT, line 122: L125: I noticed  my mistake from the past. The line slices not enough text.

The logger output is (there is a double space in the `progress  --default-path` string):

```
 [knapsack_pro] bundle exec rspec --order defined --format progress  --default-path spec "spec/controllers/pending_controller_spec.rb"
```

We could fix that.

```ruby
          stringify_cli_args = cli_args.join(' ')
          # there is a space at the end of the string
          stringify_cli_args.slice!("--format #{KnapsackPro::Formatters::RSpecQueueSummaryFormatter} ")
```

so the output would look like this:

```
 INFO -- : [knapsack_pro] bundle exec rspec --order defined --format progress --default-path spec "spec/bar_spec.rb"
```

lib/knapsack_pro/runners/queue/rspec_runner.rb
ArturT: Does it mean we don't display `--seed 123` in the very last logger message on purpose, right? I checked that, and I don't see the seed:

```
INFO -- : [knapsack_pro] To retry all the tests assigned to this CI node, please run the following command on your machine:
INFO -- : [knapsack_pro] bundle exec rspec --order random --format progress  --default-path spec "spec/pending_spec.rb" "spec/controllers/dashboard/pending_controller_spec.rb"
```

The solution that comes to my mind is to store the used seed on the first call of the `args_with_seed_option_added_when_viable` method. We could store the seed in the Ruby class variable: `@@seed`. The seed is unchanged for each batch of tests fetched from the Queue API, so it's safe to cache it in the class variable. This way, we could display the seed in the last logger message without a need to instantiate the `rspec_runner` variable which you use in L90:

```ruby
# L90
            printable_args = args_with_seed_option_added_when_viable(args, rspec_runner)
            log_rspec_command(printable_args, test_file_paths, :subset_queue)
```

ArturT (approved):
Looks good. I left some minor comments on what could be improved. It's up to you.

While I was manually testing these changes, I added bin scripts that I was using for testing. Here is related PR: https://github.com/KnapsackPro/rails-app-with-knapsack_pro/pull/40

lib/knapsack_pro/runners/queue/rspec_runner.rb
shadre: Thanks for the idea. I've updated the code to use the class variable.

While I'm not in love with the final solution, I figured it's the best one we came up with _and_ it improves the DX so I guess some trade-offs are justified.

spec/knapsack_pro/adapters/rspec_adapter_spec.rb
```diff
@@ -110,6 +110,58 @@
...
+    end
+
+    context 'when some custom order is specified' do
+      let(:cli_args) { ['--order', 'some-custom-order'] }
+
+      it { expect(subject).to eq 'some-custom-order' }
+    end
+
+    context "when the seed is given with the --seed command" do
+      let(:cli_args) { ['--seed', '123456'] }
```
ArturT, line 159: This is interesting. Good catch that you covered it. I found this in the docs:

```
--seed SEED                    Equivalent of --order rand:SEED.
```

spec/knapsack_pro/runners/queue/rspec_runner_spec.rb
```diff
@@ -274,10 +291,33 @@
...
-            exitstatus: exitstatus,
-          })
+          expect(logger).to receive(:info)
+            .with('To retry all the tests assigned to this CI node, please run the following command on your machine:')
+          expect(logger).to receive(:info).with(logged_rspec_command_matcher)
+        end
+
+        context 'when @@used_seed has been set' do
+          let(:used_seed) { '8333' }
+          let(:logged_rspec_command_matcher) { /#{args.join(' ')} --seed #{used_seed} \"#{all_test_file_paths.join(' ')}\"/ }
```
ArturT, line 301: Sometimes I wonder what to do in a situation like this.
It's easier to use programming in the spec to generate the expected string. I often do it because I like the convenience. But on the other hand, we can make mistakes in the spec as well. Who is going to test the spec itself? - once, a friend asked me :)

I had to put some mental effort into figuring out what we actually expected as a string here. It wasn't clear from reading the code.

I see one potential problem. For now, we assume `let(:all_test_file_paths) { ['a_spec.rb'] }` has one spec only.
If there would be more specs, then the `\"#{all_test_file_paths.join(' ')}\"` code does not generate a string that we actually should expect. Each spec file should be inside of apostrophes. For example, for 2 specs it should be `"a_spec.rb" "b_spec.rb"`. It's needed to ensure RSpec can run spec files from a directory with a space in the name.
Currently your coude would generated incorrect string `"a_spec.rb b_spec.rb"`.

You may consider using the expected string explicitly here so that reading the code is easier.
But I'm leaving it up to you. Whatever you prefer, it's fine with me.

ArturT (approved):
I left two potential improvement suggestions. @shadre Up to you.

spec/knapsack_pro/adapters/rspec_adapter_spec.rb
```diff
@@ -110,6 +110,58 @@
...
+      end
+    end
+
+    context "when order is 'random'" do
+      let(:cli_args) { ['--order', 'random'] }
+
+      it { expect(subject).to eq 'random' }
+
+      context 'with the seed' do
+        let(:cli_args) { ['--order', 'rand:123456'] }
```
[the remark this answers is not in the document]
shadre: Good catch, thanks

spec/knapsack_pro/runners/queue/rspec_runner_spec.rb
shadre: I agree. I'm changing this to simply include a literal path in the expectation.

PR #182, opened 2022-11-22
Add an attempt to read from the cache for Regular Mode API

ArturT:
Add an attempt to read from the cache for Regular Mode API.

This way, we can read the test suite split from the cache, and there is no need to send `test_files` in the request payload from the knapsack_pro gem for all parallel CI nodes.
This allows using Regular Mode with a large test suite (many thousands of test files across hundreds of parallel CI nodes) because requests from parallel CI nodes are fast. There is a limited number of slow requests with a big list of test files in the request payload.

# related PRs

* https://github.com/KnapsackPro/rails-app-with-knapsack_pro/pull/43
* https://github.com/KnapsackPro/docs.knapsackpro.com/pull/94

# related old PR

This PR is inspired by the improvement made for Queue Mode:

* https://github.com/KnapsackPro/knapsack_pro-ruby/pull/133
  * follow up improvement https://github.com/KnapsackPro/knapsack_pro-ruby/pull/135

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Change Log
 
+### 3.6.0
+
+* Add an attempt to read from the cache for Regular Mode API
+
+    https://github.com/KnapsackPro/knapsack_pro-ruby/pull/182
+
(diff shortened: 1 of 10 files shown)


shadre (changes requested):
Details in the PR for the API.

PR #183, opened 2022-12-02
Add before_subset_queue hook in queue mode

tubaxenor:
Currently there is only `after_subset_queue` hook and only works when subset finished. Adding a before subset hook to provide an ability to access to the subset queue to preserve some information beforehand which might be loss while CI node terminated in the middle of processing.

### DIFF
--- a/lib/knapsack_pro/hooks/queue.rb
+++ b/lib/knapsack_pro/hooks/queue.rb
@@ -3,11 +3,16 @@ module KnapsackPro
     class Queue
       class << self
         attr_reader :before_queue_store,
+          :before_subset_queue_store,
           :after_subset_queue_store,
           :after_queue_store
 
         def reset_before_queue
           @before_queue_store = nil
+        end
+
+        def reset_before_subset_queue
+          @reset_before_subset_queue = nil
         end
 
         def reset_after_subset_queue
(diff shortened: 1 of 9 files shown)


ArturT:
Hi @tubaxenor 

Please share more context why this is needed in your case.

Do you use RSpec or another test runner?

In the case of RSpec, you can run `before(:suite)` hook. It's executed before each subset queue (batch of tests fetched from Queue API).

Have you tried it? If so, why it's not good enough?
It would help me better understand your use case. Thank you.

ngan:
Hey @ArturT, you’re right, `before(:suite)` would do same thing. However, it feels really weird use the after subset hook (a Knapsack pro concept) for some things and `before(:suite)` (an RSpec concept) for others. It also seems natural to offer a before subset hook if you offer an after subset hook, no?

lib/knapsack_pro/hooks/queue.rb
```diff
@@ -3,13 +3,18 @@ module Hooks
...
+          :before_subset_queue_store,
           :after_subset_queue_store,
           :after_queue_store
 
         def reset_before_queue
           @before_queue_store = nil
         end
 
+        def reset_before_subset_queue
+          @before_subset_queue_store = nil
```
ArturT, line 15: I noticed specs where flaky (randomly failing) and I could not figure out why. It turned out there was a wrong name for this instance variable. I've fixed it and now specs are passing.

ArturT:
# code example

You can use the `before_subset_queue` hook to preview list of test files fetched from the Queue API in a given batch of tests.

```ruby
KnapsackPro::Hooks::Queue.before_subset_queue do |queue_id|
  # list of test files fetched from the Queue API for a given batch/subset
  # tests are not executed yet so the time_execution is 0 seconds
  puts KnapsackPro.tracker.to_a
end
```

I've released the knapsack_pro gem 3.5.0 version.

PR #184, opened 2022-12-28
Adjust the timer behaviour in the RSpec adapter

shadre:
Prevent the timer being stopped in the same test file.

Shout-out to @evanlok for the inspiration and code example

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Change Log
 
+### 3.7.0
+
+* Adjust the timer behaviour in the RSpec adapter
+
+    https://github.com/KnapsackPro/knapsack_pro-ruby/pull/184
+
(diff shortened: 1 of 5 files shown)


ArturT (approved):
It seems to work fine. 👍 I tested it locally with [a rails app project by commenting out this line](https://github.com/KnapsackPro/rails-app-with-knapsack_pro/blob/e1a9fbcf9d7081ef5901076751b40c6ce2bead86/bin/edge_cases/knapsack_pro_queue_rspec_in_background#L18).


You could update the changelog as part of this PR.

ArturT:
> Prevent the timer being stopped in the same test file.

Was this also causing some of the test files to have 0 seconds of execution time? I was not sure how to reproduce this edge case.

shadre:
> Was this also causing some of the test files to have 0 seconds of execution time? I was not sure how to reproduce this edge case.

It seems plausible but I am not entirely sure about that.

PR #186, opened 2023-02-10
Remove from Readme what was moved to docs

3v0k4:
And make some improvements on the way


ArturT (approved):
Approved. The next step could be to update links that we print in the output when running tests with the knapsack_pro gem. Some of them leads to old FAQ or to the old readme. Grep repo by:

* github.com/KnapsackPro/knapsack_pro-ruby
* knapsackpro.com/faq

PR #187, opened 2023-02-23
🚧 WIP: Extract links

3v0k4:
`rg "https://" lib`


spec/integration/urls_spec.rb
```diff
@@ -0,0 +1,40 @@
+describe 'Permalinks work' do
```
3v0k4, line 1: How do you feel about this test (that actually hits https://knapsackpro.com)?

shadre: My take is that the redirect behavior should be tested on the API, not in the library. Same way we don't test other API endpoints from the library. IMO the API sets the standard (permalinks), and the library should just follow it. In essence, I view the permalinks as an internal API of sorts.

ArturT (commented):
It looks good.

I added comments for [the PR on the API side](https://github.com/KnapsackPro/knapsack-pro-api/pull/630/) that could impact the links here.

Issue #189, opened 2023-03-21
Regular mode (non-queue) rspec outputs all spec file names on startup


alex-kovshovik:
We (Decisely) use knapsack_pro gem for a long time (years now), and it's been **solid**! We never had issues with KnapsackPro service being down, never had problems with running older versions of the gem. This is the best open-source solution to spread our large test suite across nodes in our GitLab CI. Thank you!

The one issue that's not really a bug, it's just an annoyance: we have a lot of spec files, and when KnutsackPro runs them using `rake spec` - it outputs all these file names to STDOUT, which in our case is several pages out output. So whenever a test fails in CI, a developer must scroll a lot to find it in the CI output.

I changed the KnapsackPro logging level to "warn" with `KNAPSACK_PRO_LOG_LEVEL: "warn"` and it helped, but still `bundle exec rails knapsack_pro:rspec` prints out ALL file names.

**Question:** How can I suppress the output of the file names?

Example output:

> $ bundle exec rails knapsack_pro:rspec
[...]

PR #190, opened 2023-03-21
Suppress all RSpec spec file names displayed in stdout at the beginning of running tests in Regular Mode only when the log level is >= `warn`

ArturT:
In Regular Mode, we use RSpec::Core::RakeTask. By default, it prints an RSpec command with all test files and options provided to it. For a big test suite, this can be a large output.
We can suppress that with a verbose false option for the rake task only when [the log level is >= `warn`](https://docs.knapsackpro.com/ruby/reference/#knapsack_pro_log_level). 

# related issue

* https://github.com/KnapsackPro/knapsack_pro-ruby/issues/189

# related

https://stackoverflow.com/a/26299828/905697

### DIFF
--- a/lib/knapsack_pro/runners/rspec_runner.rb
+++ b/lib/knapsack_pro/runners/rspec_runner.rb
@@ -32,6 +32,7 @@ module KnapsackPro
             # instead we pass test files and test example paths to t.rspec_opts
             t.pattern = []
             t.rspec_opts = "#{args} --default-path #{runner.test_dir} #{runner.stringify_test_file_paths}"
+            t.verbose = KnapsackPro::Config::Env.log_level < ::Logger::WARN
           end
           Rake::Task[task_name].invoke
         end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Change Log
 
+### 3.9.0
+
+* Suppress all RSpec spec file names displayed in stdout at the beginning of running tests in Regular Mode only when the log level is >= `warn`
+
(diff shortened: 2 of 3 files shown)


lib/knapsack_pro/runners/rspec_runner.rb
```diff
@@ -32,6 +32,7 @@ def self.run(args)
             # instead we pass test files and test example paths to t.rspec_opts
             t.pattern = []
             t.rspec_opts = "#{args} --default-path #{runner.test_dir} #{runner.stringify_test_file_paths}"
+            t.verbose = false
```
ArturT, line 35: I'm wondering if this should be an env var so that the user could decide to suppress the output.

If we make it `verbose=false` as a default, then this suggestion here:

```
You can also run the same subset of tests without Knapsack Pro: in the logs, find the command that Knapsack Pro used to invoke the test runner:

rspec  --default-path spec "spec/models/user_spec.rb" "spec/models/comment_spec.rb"
```

https://docs.knapsackpro.com/ruby/troubleshooting/#regular-mode

It won't be possible anymore to copy&paste RSpec command to run tests locally.

ArturT: I modified the code. It will be the most user-friendly to respect the log level. If someone uses >= warn log level, then we suppress the RSpec output with a list of test files to execute.

shadre: I like the conclusion you came to

shadre (approved):

PR #195, opened 2023-05-17
feat: KNAPSACK_PRO_CI_NODE_BUILD_ID required

3v0k4:


spec/knapsack_pro/config/env_spec.rb
```diff
@@ -72,8 +72,10 @@
       end
     end
 
-    context "when ENV doesn't exist" do
-      it { should eq 'missing-build-id' }
+    context "when ENV does not exist" do
+      it 'raises' do
+        expect { subject }.to raise_error
```
ArturT, line 77:
```suggestion
        expect { subject }.to raise_error(/Missing environment variable KNAPSACK_PRO_CI_NODE_BUILD_ID/)
```

lib/knapsack_pro/config/env.rb
```diff
@@ -23,9 +23,10 @@ def ci_node_index
         end
 
         def ci_node_build_id
-          ENV['KNAPSACK_PRO_CI_NODE_BUILD_ID'] ||
+          env_name = 'KNAPSACK_PRO_CI_NODE_BUILD_ID'
+          ENV[env_name] ||
             ci_env_for(:node_build_id) ||
-            'missing-build-id'
+            raise("Missing environment variable #{env_name}. Read more at #{KnapsackPro::Urls::KNAPSACK_PRO_CI_NODE_BUILD_ID}")
```
ArturT, line 29: You could update [CHANGLELOG](https://github.com/KnapsackPro/knapsack_pro-ruby/blob/master/CHANGELOG.md) as part of this PR. Please also make sure you release a major version 4.0.0 because we have a breaking change here. If someone uses CI provider without out-of-the-box CI build ID support then we could break their CI builds with this change when they upgrade the gem.

ArturT (approved):

PR #199, opened 2023-05-25
Set `RAILS_ENV=test`/`RACK_ENV=test` in Queue Mode

3v0k4:
We want to make Queue Mode as easy as changing

```bash
bundle exec knapsack_pro:RUNNER
```

to

```bash
bundle exec knapsack_pro:queue:RUNNER
```

This PR removes the need to set `RAILS_ENV=test` like

```bash
RAILS_ENV=test bundle exec knapsack_pro:queue:RUNNER
```

---

I tried to set `RAILS_ENV=test` and `RACK_ENV=test` in two different places
- right before calling the runner in the rake file
- in the runner

but I got some (strange) errors when running `bin/knapsack_pro_all`.

In particular, the following error popped out to me (to reproduce remove `  RAILS_ENV=test \
` from `bin/knapsack_pro_queue_minitest` and run `bin/knapsack_pro_queue_minitest`):

```
Error:
ArticlesControllerTest#test_should_get_index:
RuntimeError: Neutered Exception NoMethodError: assigns has been extracted to a gem. To continue using it,
        add `gem 'rails-controller-testing'` to your Gemfile.
    test/controllers/articles_controller_test.rb:11:in `block in <class:ArticlesControllerTest>'
```

The `rails-controller-testing` gem is in the `test` group, which made me think the app needs to be loaded WITH `RAILS_ENV=test`. So I decided to try by shelling out to a separate process by using `sh`. That did the trick and `bin/knapsack_pro_all` passes.

### DIFF
--- a/lib/knapsack_pro/runners/queue/cucumber_runner.rb
+++ b/lib/knapsack_pro/runners/queue/cucumber_runner.rb
@@ -3,6 +3,8 @@ module KnapsackPro
     module Queue
       class CucumberRunner < BaseRunner
         def self.run(args)
+          # ENV['RAILS_ENV'] = ENV['RACK_ENV'] = 'test'
+
           require 'cucumber/rake/task'
 
           ENV['KNAPSACK_PRO_TEST_SUITE_TOKEN'] = KnapsackPro::Config::Env.test_suite_token_cucumber
--- a/lib/knapsack_pro/runners/queue/cucumber_runner.rb
+++ b/lib/knapsack_pro/runners/queue/cucumber_runner.rb
@@ -3,8 +3,6 @@ module KnapsackPro
     module Queue
       class CucumberRunner < BaseRunner
         def self.run(args)
-          # ENV['RAILS_ENV'] = ENV['RACK_ENV'] = 'test'
-
           require 'cucumber/rake/task'
 
           ENV['KNAPSACK_PRO_TEST_SUITE_TOKEN'] = KnapsackPro::Config::Env.test_suite_token_cucumber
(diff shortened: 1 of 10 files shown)


spec/fixtures/vcr_cassettes/api/v1/build_distributions/subset/timeout.yml
```diff
@@ -0,0 +1,106 @@
+---
```
ArturT, line 1: I believe no cassettes at all should be recorded when the domain is fake. You could kill your http server running on 3000 port before you run tests.

3v0k4:
I researched more why 

> The rails-controller-testing gem is in the test group, which made me think the app needs to be loaded WITH RAILS_ENV=test. So I decided to try by shelling out to a separate process by using sh. That did the trick and bin/knapsack_pro_all passes.

When invoking `bundle exec rake knapsack_pro:...` this is the call order:

- `rake` (in application.rb) loads the Rakefile
- the Rakefile requires config/application.rb
- config/application.rb calls Bundler.require(*Rails.groups)
- `Rails.groups` returns `[:default, :development]` so the the development gems are not required

lib/knapsack_pro/runners/queue/cucumber_runner.rb
```diff
@@ -3,6 +3,8 @@ module Runners
     module Queue
       class CucumberRunner < BaseRunner
         def self.run(args)
+          # ENV['RAILS_ENV'] = ENV['RACK_ENV'] = 'test'
```
ArturT, line 6: We are good to remove comments now since we establish solution with `$PROGRAM_NAME`.

ArturT (approved):

ArturT (approved):
I think we are good to release it.

PR #202, opened 2023-06-08
Mask user seats data instead of hashing it

shadre:


ArturT (approved):
Looks good. Please note we hash user seat value again on the API side so we would have to remove hashing on the API side.

PR #208, opened 2023-07-06
Send authors to API

3v0k4:
Send masked authors to API:
- `build_author` is the name and email of the author for the last commit
- `commit_authors` is author (as above) + number of commits obtained by grouping all the commits of the last month

A `git fetch` is performed to make sure the last months of commits are available in case the repository is performing shallow clones for performance reasons.

Notice that this solution respects [gitmailmap](https://git-scm.com/docs/gitmailmap), which could be useful to customers/us.

See also https://github.com/KnapsackPro/knapsack-pro-api/pull/726


lib/knapsack_pro/repository_adapters/git_adapter.rb
```diff
@@ -14,8 +14,34 @@ def branches
...
 
+      def authors
+        git_authors
+          .split("\n")
+          .map { |line| line.strip }
+          .map { |line| line.split("\t") }
+          .map do |commits, author|
+            { commits: commits.to_i, author: KnapsackPro::MaskString.call(author) }
+          end
+      rescue Exception
```
3v0k4, line 25: I thought about just `rescue`ing the git command but I wanted to go 100% safe

shadre (approved):
Looks Neat To me 🚀

spec/knapsack_pro/repository_adapters/git_adapter_spec.rb
```diff
@@ -31,4 +31,56 @@
...
+
+        expect(subject.author).to eq 'no git <no.git@example.com>'
+      end
+    end
+  end
+
+  describe '#authors' do
+    it "returns the masked authors" do
+      allow_any_instance_of(KnapsackPro::RepositoryAdapters::GitAdapter).to receive(:git_authors).and_return([
+        "     5\t3v0k4 <riccardo.odone@gmail.com>\n",
```
ArturT, line 60: We could use fake emails here instead of real personal data in a public open source project.

lib/knapsack_pro/repository_adapters/git_adapter.rb
ArturT: Ok.

I tested using a fake command `xgit`:

```
3.2.2 :005 >  def git_authors
3.2.2 :006 >         `xgit shortlog --summary --email --since "one month ago"`
3.2.2 :007 >  end


3.2.2 :008 > git_authors
sh: xgit: command not found
 => ""

3.2.2 :015 > git_authors
3.2.2 :016 >           .split("\n")
3.2.2 :017 >           .map { |line| line.strip }
3.2.2 :018 >           .map { |line| line.split("\t") }
3.2.2 :019 >           .map do |commits, author|
3.2.2 :020?>             { commits: commits.to_i, author: KnapsackPro::MaskString.call(author)
}
3.2.2 :021 >           end
sh: xgit: command not found
 => []
```

Is it fine to return an empty array here? Maybe it would be better to return:

```
[{ commits: 0, author: "no git <no.git@example.com>" }]
```

lib/knapsack_pro/repository_adapters/git_adapter.rb
```diff
@@ -14,8 +14,34 @@ def branches
...
+        [{ commits: 0, author: "no git <no.git@example.com>" }]
+      end
+
+      def author
+        KnapsackPro::MaskString.call(git_author.strip)
```
ArturT, line 30: [A similar problem](https://github.com/KnapsackPro/knapsack_pro-ruby/pull/208/files#r1261027819) is for the `git_author` method. It could return an empty string when git command is not found in the system.

Maybe for the empty string (git_author) we should return `"no git <no.git@example.com>"`.

ArturT (approved):
I left a few ideas on what we could improve.

lib/knapsack_pro/repository_adapters/git_adapter.rb
3v0k4: I'm not sure I follow. The entire body of `authors` is rescued, so that would always return `[{ commits: 0, author: "no git <no.git@example.com>" }]` in case of an exception.

Also, `git_authors` is not public and it's only called by `authors`.

3v0k4: I'm not sure I follow.

The entire body of authors is rescued, so that would always return [{ commits: 0, author: "no git <no.git@example.com>" }] in case of an exception.

Also, git_authors is not public and it's only called by authors.

Example:

```bash
➜ irb
3.2.2 :001 > def cmd; `xgit`; end
 => :cmd 
3.2.2 :002 > def go
3.2.2 :003 >   cmd
3.2.2 :004 > rescue Exception
3.2.2 :005 >   "rescued"
3.2.2 :006 > end
 => :go 
3.2.2 :007 > go
 => "rescued" 

```

lib/knapsack_pro/repository_adapters/git_adapter.rb
3v0k4: See comment above

lib/knapsack_pro/repository_adapters/git_adapter.rb
ArturT: Hmm. When I use the above example with the `cmd` method I also get the exception but when I use this:

```
def git_authors
  `xgit shortlog --summary --email --since "one month ago"`
end
```

then there is no exception raised. Only empty string is returned and the `sh: xgit: command not found` message is printed to the output.

```
3.2.2 :008 > git_authors
sh: xgit: command not found
 => ""
```

My main point is that we should also cover case when the method returns an empty string because catching exception seems to not be enough.

ArturT (approved):

lib/knapsack_pro/client/api/v1/queues.rb
```diff
@@ -17,6 +17,8 @@ def queue(args)
                 :node_index => args.fetch(:node_index),
                 :node_build_id => KnapsackPro::Config::Env.ci_node_build_id,
                 :user_seat => KnapsackPro::Config::Env.masked_user_seat,
+                :build_author => KnapsackPro::RepositoryAdapters::GitAdapter.new.build_author,
+                :commit_authors => KnapsackPro::RepositoryAdapters::GitAdapter.new.commit_authors,
```
ArturT, line 21: We could do a performance improvement here and move the lines down below. I completely forgot about it. Just noticed it now.

We can send build author and commit authors only in the request that supposed to initialize a new queue. Thanks to that we don't have to transfer unnecessary data over the network for each Queue API request that is consuming tests from the Queue. Similarly like we do for test files. 

Additional benefit of that approach is that we don't have to call git command on each request. I guess this could cause a small delay in running tests if you call git many times on each CI node for each request.





```ruby
            if request_hash[:can_initialize_queue] && !request_hash[:attempt_connect_to_queue]
                request_hash.merge!({
                  :test_files => args.fetch(:test_files),
                  :build_author => KnapsackPro::RepositoryAdapters::GitAdapter.new.build_author,
                  :commit_authors => KnapsackPro::RepositoryAdapters::GitAdapter.new.commit_authors
                })
            end
```

lib/knapsack_pro/repository_adapters/git_adapter.rb
```diff
@@ -14,8 +14,40 @@ def branches
...
+        raise if author.empty?
+        author
+      rescue Exception
+        "no git <no.git@example.com>"
+      end
+
       private
 
+      def git_commit_authors
+        `git fetch --shallow-since "1 month ago" >/dev/null 2>&1 && git shortlog --summary --email --since "one month ago"`
```
ArturT, line 44: If you run knapsack pro locally on your machine, you call `--shallow-since` command and when I want to use `git blame`, I only see commits from the last month.

I undo changes by using the following command:

```
git fetch --unshallow || git fetch --all
```

I guess someone might not like that we mess up with their local repository when they try to run Knapsack Pro locally.

3v0k4: Thanks for the report.

ArturT: For reference: The above issue will be solved when we run `--shallow-since` only in the context of the CI. This PR solves it:

* https://github.com/KnapsackPro/knapsack_pro-ruby/pull/209

PR #209, opened 2023-07-18
Improve commit authors tracking

3v0k4:
Tested on:
- CircleCI
- Buildkite
  - this worked also with the previous code
- Github Actions
  - this **did not** work with the previous code
  
More info on the fix: https://stackoverflow.com/questions/73085141/git-shortlog-in-a-github-workflow-for-a-specific-directory

### DIFF
--- a/lib/knapsack_pro/repository_adapters/git_adapter.rb
+++ b/lib/knapsack_pro/repository_adapters/git_adapter.rb
@@ -41,7 +41,11 @@ module KnapsackPro
       private
 
       def git_commit_authors
-        `git fetch --shallow-since "1 month ago" --quiet && git shortlog --summary --email --since "one month ago"`
+        if KnapsackPro::Config::Env.ci?
+          `git fetch --shallow-since "one month ago" --quiet`
+        end
+
+        `git shortlog --summary --email --since "one month ago"`
       end
 
       def git_build_author
--- a/lib/knapsack_pro/repository_adapters/git_adapter.rb
+++ b/lib/knapsack_pro/repository_adapters/git_adapter.rb
@@ -41,7 +41,7 @@ module KnapsackPro
       private
 
       def git_commit_authors
-        `git fetch --shallow-since "1 month ago" >/dev/null 2>&1 && git shortlog --summary --email --since "one month ago"`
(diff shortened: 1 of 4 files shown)


lib/knapsack_pro/repository_adapters/git_adapter.rb
```diff
@@ -41,7 +42,11 @@ def build_author
       private
 
       def git_commit_authors
-        `git fetch --shallow-since "1 month ago" >/dev/null 2>&1 && git shortlog --summary --email --since "one month ago"`
+        if ENV.fetch('CI', 'false') == 'true'
```
ArturT, line 45:
```suggestion
        if ENV.fetch('CI', 'false') == 'true' || KnapsackPro::Config::Env.detected_ci
```

Nice idea with checking CI=true here. Additionally we could check if any CI server was detected by us with the `detected_ci` method.
https://github.com/KnapsackPro/knapsack_pro-ruby/blob/f667808c9f02f918fcadd5f535bbde76f67be383/lib/knapsack_pro/config/env.rb#L262
https://github.com/KnapsackPro/knapsack_pro-ruby/blob/f667808c9f02f918fcadd5f535bbde76f67be383/lib/knapsack_pro/config/ci/base.rb#L29
I'm not sure if all CI providers use the `CI` env variable.

lib/knapsack_pro/repository_adapters/git_adapter.rb
```diff
@@ -16,6 +16,7 @@ def branches
 
       def commit_authors
         authors = git_commit_authors
+          .yield_self { |res| puts res.inspect; res }
```
ArturT, line 19: I guess we can remove `puts` before merging this.

lib/knapsack_pro/repository_adapters/git_adapter.rb
ArturT: We can also use `.downcase` method because some CI providers use `True` as value.
https://www.appveyor.com/docs/environment-variables/

```ruby
ENV.fetch('CI', 'false').downcase == 'true'
```

ArturT (approved):
I added some suggestions on what we could improve.

ArturT (approved):

PR #216, opened 2023-08-16
Send CI provider via header

3v0k4:
Related: https://github.com/KnapsackPro/knapsack-pro-api/pull/754

`ci_provider` is the same as the ones users are [filling in the first step of onboarding](https://github.com/KnapsackPro/knapsack-pro-api/blob/7791c04a7f0593b32e44e32c4f42aa34eb4dd3f3/app/form_objects/onboarding/project_form.rb#L17
) minus the following that currently are marked as "Other":

- Argo CD
- Concourse
- Flux CD
- GoCD
- Harness.io
- Jenkins X
- Prow
- Tekton

I made sure we cover at least the top CIs from the dashboard.

Also, I took some inspiration from https://github.com/watson/ci-info/blob/master/vendors.json for the additional checks in `Config::CI::Base`.


lib/knapsack_pro/config/ci/base.rb
```diff
@@ -32,6 +32,17 @@ def detected
...
+
+        def ci_provider
+          return 'AWS CodeBuild' if ENV.key?('CODEBUILD_BUILD_ARN')
+          return 'Azure Pipelines' if ENV.key?('SYSTEM_TEAMFOUNDATIONCOLLECTIONURI')
+          return 'Bitbucket Pipelines' if ENV.key?('BITBUCKET_COMMIT')
+          return 'Google Cloud Build' if ENV.key?('BUILDER_OUTPUT')
+          return 'Jenkins' if ENV.key?('JENKINS_URL')
+          return 'TeamCity' if ENV.key?('TEAMCITY_VERSION')
+
+          "Other"
```
ArturT, line 44:
```suggestion
          "Other" if ENV.key?('CI')
```

Maybe we should do something like that. 

Thanks to that if you are running knapsack inside of the CI then it's `Other` (not perfect solution because we can't probably assume each CI has `CI=true` value set. But most do it so maybe that's the good enough assumption).
If you run it in the development then we won't override the existing detected CI on the API side by the `Other` value.

ArturT (approved):

lib/knapsack_pro/config/ci/base.rb
```diff
@@ -32,6 +32,17 @@ def detected
...
         end
+
+        def ci_provider
+          return 'AWS CodeBuild' if ENV.key?('CODEBUILD_BUILD_ARN')
+          return 'Azure Pipelines' if ENV.key?('SYSTEM_TEAMFOUNDATIONCOLLECTIONURI')
+          return 'Bitbucket Pipelines' if ENV.key?('BITBUCKET_COMMIT')
+          return 'Google Cloud Build' if ENV.key?('BUILDER_OUTPUT')
+          return 'Jenkins' if ENV.key?('JENKINS_URL')
+          return 'TeamCity' if ENV.key?('TEAMCITY_VERSION')
+
```
ArturT, line 43: We could consider detecting also Buddy Works, Drone, Bamboo. I recall some users were using it.

3v0k4: https://github.com/KnapsackPro/knapsack_pro-ruby/pull/216/commits/f05d807e7b772eaeb01d99ddb7bca44810d9442b

lib/knapsack_pro/config/ci/base.rb
3v0k4: https://github.com/KnapsackPro/knapsack_pro-ruby/pull/216/commits/3c3881f885641c0424fbfced38ca2186cac422d3

3v0k4:
Looks good to release?

lib/knapsack_pro/client/connection.rb
```diff
@@ -59,18 +59,15 @@ def request_body
...
         {
           'Content-Type' => 'application/json',
           'Accept' => 'application/json',
           'KNAPSACK-PRO-CLIENT-NAME' => client_name,
           'KNAPSACK-PRO-CLIENT-VERSION' => KnapsackPro::VERSION,
-          'KNAPSACK-PRO-TEST-SUITE-TOKEN' => test_suite_token,
-        }
+          'KNAPSACK-PRO-TEST-SUITE-TOKEN' => KnapsackPro::Config::Env.test_suite_token,
+          'KNAPSACK-PRO-CI-PROVIDER' => KnapsackPro::Config::Env.ci_provider,
+        }.compact
```
ArturT, line 70: Since you added `compact`. Suppose someone forgot to set the API token. Will the API still return a meaningful error?

ArturT (approved):
Just one small thing to check https://github.com/KnapsackPro/knapsack_pro-ruby/pull/216#discussion_r1298602445
and it looks good to release.

lib/knapsack_pro/client/connection.rb
3v0k4: @ArturT I removed the token from the script and I got this:

```bash
bin/knapsack_pro_rspec  
rake aborted!
Missing environment variable KNAPSACK_PRO_TEST_SUITE_TOKEN. You should set environment variable like KNAPSACK_PRO_TEST_SUITE_TOKEN_RSPEC (note there is suffix _RSPEC 
at the end). knapsack_pro gem will set KNAPSACK_PRO_TEST_SUITE_TOKEN based on KNAPSACK_PRO_TEST_SUITE_TOKEN_RSPEC value. If you use other test runner than RSpec then 
use proper suffix.
```

It works because we first build the hash calling `KnapsackPro::Config::Env.test_suite_token` and, only then, `compact` it.

ArturT: ok. Thanks for checking.

Issue #224, opened 2023-09-07
Slow test analysis runs even when there's no slow tests


ngan:
We don't always build every single test but it looks like the slow test analysis always runs even when there's no slow test:
https://github.com/KnapsackPro/knapsack_pro-ruby/blob/master/lib/knapsack_pro/base_allocator_builder.rb#L55

No slow tests being that [`slow_test_files = get_slow_test_files`](https://github.com/KnapsackPro/knapsack_pro-ruby/blob/master/lib/knapsack_pro/base_allocator_builder.rb#L44C17-L44C17) is an empty array.

An optimization to make would be to see if we actually got any slow tests back from the `get_slow_test_files` call. If we got an empty array, it should just bypass the [analysis](https://github.com/KnapsackPro/knapsack_pro-ruby/blob/master/lib/knapsack_pro/base_allocator_builder.rb#L46-L61) completely.

shadre:
Thanks for the suggestion.

We are aware of that and plan to address this. I'm closing this as it's a dup of: https://github.com/KnapsackPro/knapsack_pro-ruby/issues/223.

ArturT:
@ngan This is solved in the knapsack_pro gem [5.7.0 version](https://github.com/KnapsackPro/knapsack_pro-ruby/blob/master/CHANGELOG.md#570). Please update the knapsack_pro gem in your project.

Issue #223, opened 2023-08-30
Does `rake knapsack_pro:rspec_test_example_detector` need to run if there aren't any slow specs?


technicalpickles:
While doing some local testing, I have sometimes seen the log output from this line:

https://github.com/KnapsackPro/knapsack_pro-ruby/blob/4d9dc1436470995773aa080d8054f6c587aef91b/lib/knapsack_pro/base_allocator_builder.rb#L46

Does the rake task still need to run if  `slow_test_files.size == 0`? It is relatively fast (15s locally), but if there aren't any slow specs, it'd be nice to save a little time.

ArturT:
I guess it could be skipped, then. We'll have to check that.

ArturT:
@technicalpickles This is solved in the knapsack_pro gem [5.7.0 version](https://github.com/KnapsackPro/knapsack_pro-ruby/blob/master/CHANGELOG.md#570). Please update the knapsack_pro gem in your project.

PR #225, opened 2023-09-19
Performance improvement: don't run `rake knapsack_pro:rspec_test_example_detector` when no slow test files are detected for RSpec

ArturT:
Performance improvement: don't run `rake knapsack_pro:rspec_test_example_detector` when no slow test files are detected for RSpec. 

Thanks to that, we improve performance and save a few or a dozen seconds (depending on the project's complexity).

# Related issues

* https://github.com/KnapsackPro/knapsack_pro-ruby/issues/223
* https://github.com/KnapsackPro/knapsack_pro-ruby/issues/224

# development testing

This PR change was tested with a [bin script](https://github.com/KnapsackPro/rails-app-with-knapsack_pro/blob/0e84152883167e4bb197ebc4c05bb9bde5e3b811/bin/knapsack_pro_queue_rspec_split_by_test_examples_unique_build#L5).

### DIFF
--- a/lib/knapsack_pro/base_allocator_builder.rb
+++ b/lib/knapsack_pro/base_allocator_builder.rb
@@ -35,18 +35,13 @@ module KnapsackPro
     def fast_and_slow_test_files_to_run
       test_files_to_run = all_test_files_to_run
 
-      if adapter_class == KnapsackPro::Adapters::RSpecAdapter && KnapsackPro::Config::Env.rspec_split_by_test_examples?
-        require 'rspec/core/version'
-        unless Gem::Version.new(::RSpec::Core::Version::STRING) >= Gem::Version.new('3.3.0')
-          raise "RSpec >= 3.3.0 is required to split test files by test examples. Learn more: #{KnapsackPro::Urls::SPLIT_BY_TEST_EXAMPLES}"
-        end
-
+      if adapter_class.split_by_test_cases_enabled?
         slow_test_files = get_slow_test_files
         return test_files_to_run if slow_test_files.empty?
 
-        test_file_example_paths = rspec_test_file_example_paths_for_slow_test_files(slow_test_files)
+        test_file_cases = adapter_class.test_file_cases_for(slow_test_files)
 
-        return KnapsackPro::TestFilesWithTestCasesComposer.call(test_files_to_run, slow_test_files, test_file_example_paths)
+        return KnapsackPro::TestFilesWithTestCasesComposer.call(test_files_to_run, slow_test_files, test_file_cases)
       end
 
       test_files_to_run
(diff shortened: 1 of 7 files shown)


lib/knapsack_pro/base_allocator_builder.rb
3v0k4, line 1: Have you considered moving the logic specific to an adapter inside the `adapter_class`?

We could prolly replace most of `fast_and_slow_test_files_to_run` and its `if adapter_class == KnapsackPro::Adapters::RSpecAdapter && KnapsackPro::Config::Env.rspec_split_by_test_examples?` with something like:

```
adapter_class.new.some_name_that_makes_sense(args)
```

lib/knapsack_pro/base_allocator_builder.rb
```diff
@@ -101,5 +87,24 @@ def get_slow_test_files
...
+        KnapsackPro::Config::Env.rspec_test_example_detector_prefix,
+        'rake knapsack_pro:rspec_test_example_detector',
+      ].join(' ')
+      unless Kernel.system(cmd)
+        raise "Could not generate JSON report for RSpec. Rake task failed when running #{cmd}"
+      end
+
+      # read the JSON report
+      detector = KnapsackPro::TestCaseDetectors::RSpecTestExampleDetector.new
+      test_file_example_paths = detector.test_file_example_paths
```
3v0k4, line 107:
```suggestion
      KnapsackPro::TestCaseDetectors::RSpecTestExampleDetector
        .new
        .test_file_example_paths
```

spec/knapsack_pro/base_allocator_builder_spec.rb
```diff
@@ -154,7 +154,7 @@
       end
 
       context 'when rake task to detect RSpec test examples works' do
-        let(:slow_test_files) { double(size: 5) }
+        let(:slow_test_files) { double(size: 5, empty?: false) }
```
3v0k4, line 157: Is `slow_test_files` an array? If so, I'd pass a real thing rather than a stub.

spec/knapsack_pro/base_allocator_builder_spec.rb
3v0k4, line 1: Are we covering the `empty?: true` case?

lib/knapsack_pro/base_allocator_builder.rb
```diff
@@ -101,5 +87,24 @@ def get_slow_test_files
       KnapsackPro.logger.debug("Detected #{slow_test_files.size} slow test files: #{slow_test_files.inspect}")
       slow_test_files
     end
+
+    def rspec_test_file_example_paths_for_slow_test_files(slow_test_files)
```
3v0k4, line 91: If you move this into the adapter, I guess you could do:

```suggestion
    def example_paths_for(test_files)
```

3v0k4 (approved):

spec/knapsack_pro/base_allocator_builder_spec.rb
ArturT: Yes. It's covered now.

spec/knapsack_pro/base_allocator_builder_spec.rb
ArturT: Ok. I use a real array now.

lib/knapsack_pro/base_allocator_builder.rb
ArturT: done

lib/knapsack_pro/base_allocator_builder.rb
ArturT: done.

lib/knapsack_pro/base_allocator_builder.rb
ArturT: Good idea. I extracted everything related to the RSpec to the adapter class. Other generic things stay in base allocator builder so that we have option to extend other adapters to also support test cases splitting in the future.

PR #230, opened 2023-12-01
fix changelog link

cccCody:


ArturT:
Good catch! Thanks @cccCody

PR #231, opened 2023-12-01
fix(RSpec): allow to use Queue Mode for RSpec <= 3.10 when the `rspec_is_quitting` method is not present for RSpec World object

ArturT:
# story

https://trello.com/c/0ZISAmYr

# bug in RSpec 3.10

RSpec 3.10 has no method `rspec_is_quitting` and it fails when you use Queue Mode in the knapsack_pro gem.

RSpec 3.10 - the `rspec_is_quitting` method is missing: https://github.com/rspec/rspec-core/blob/v3.10.0/lib/rspec/core/world.rb

RSpec 3.12 - [has the method](https://github.com/rspec/rspec-core/blob/1eeadce5aa7137ead054783c31ff35cbfe9d07cc/lib/rspec/core/world.rb#L18)

# fix

Call the `rspec_is_quitting` method only if it exists on RSpec World object.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+### 6.0.1
+
+* fix(RSpec): allow to use Queue Mode for RSpec <= 3.10 when the `rspec_is_quitting` method is not present for RSpec World object
+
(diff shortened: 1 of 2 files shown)


3v0k4 (approved):
🙌

PR #233, opened 2023-12-07
Fix Turnip and UNKNOWN_PATH error

3v0k4:
Turnip specs have a different naming format than RSpec's: `foo.feature` vs `foo_spec.rb`.

This PR makes sure `.feature` files are recorded.

Also, it stops recording `UNKNOWN_PATH` that would generate an error in case of a CI node retry (see below for details).

## Reproduce
- In rails-app-with-knapsack-pro
  - `git commit --allow-empty -m wip`
  - `bin/knapsack_pro_queue_turnip 0 2 20231207-01`
  - `bin/knapsack_pro_queue_turnip 0 2 20231207-01`

You should get:

```
D, [2023-12-07T14:08:19.969425 #14329] DEBUG -- : [knapsack_pro] POST http://api.knapsackpro.test:3000/v1/queues/queue
D, [2023-12-07T14:08:19.969501 #14329] DEBUG -- : [knapsack_pro] API request UUID: 668869b5-a70d-4261-a624-3f47e31d9971
D, [2023-12-07T14:08:19.969553 #14329] DEBUG -- : [knapsack_pro] API response:
D, [2023-12-07T14:08:19.969650 #14329] DEBUG -- : [knapsack_pro] {"queue_name"=>nil, "build_subset_id"=>"24fcce5d-06b5-4f96-bcff-2e9e4aba3292", "test_files"=>[{"path
"=>"UNKNOWN_PATH", "time_execution"=>0.09200099995359778}, {"path"=>"turnip/acceptance/foo.feature", "time_execution"=>0.0}]}
D, [2023-12-07T14:08:19.995617 #14329] DEBUG -- : [knapsack_pro] Test suite time execution queue recording enabled.

An error occurred while loading ./UNKNOWN_PATH.
Failure/Error: super

LoadError:
  cannot load such file -- /Users/riccardoodone/code/knapsack/rails-app-with-knapsack_pro/UNKNOWN_PATH
# /Users/riccardoodone/.rvm/gems/ruby-3.2.2/gems/turnip-4.4.0/lib/turnip/rspec.rb:20:in `load'
# /Users/riccardoodone/.rvm/gems/ruby-3.2.2/gems/turnip-4.4.0/lib/turnip/rspec.rb:20:in `load'
# /Users/riccardoodone/code/knapsack/knapsack_pro-ruby/lib/knapsack_pro/runners/queue/rspec_runner.rb:99:in `run_tests'
[...]

PR #235, opened 2023-12-15
integration tests

3v0k4:
This PR takes care of the following:
- use CircleCI configuration v2.1
- (besides the gem tests) run integration tests by cloning rails-app-with-knapsack-pro
- remove the CodeClimate ENV because we are not using it
- make sure Rubocop exits with an error if there's an offense (and fix a small offense)

This setup catches:
- #231 (https://app.circleci.com/pipelines/github/KnapsackPro/knapsack_pro-ruby/1055/workflows/3dd23caf-13d7-4683-b11e-dd5aacf969cf/jobs/2421, https://app.circleci.com/pipelines/github/KnapsackPro/knapsack_pro-ruby/1055/workflows/3dd23caf-13d7-4683-b11e-dd5aacf969cf/jobs/2429)
- #233 (https://app.circleci.com/pipelines/github/KnapsackPro/knapsack_pro-ruby/1059/workflows/b46cd967-adad-4615-9837-ae932f1ff703)

See the builds for this branch on [CircleCI](https://app.circleci.com/pipelines/github/KnapsackPro/knapsack_pro-ruby?branch=integration-tests).

I decided not to add the next version of ruby (3.3) because it would be more complex than waiting for CircleCI to add it to their [cimg images](https://circleci.com/developer/images/image/cimg/ruby).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -76,6 +76,7 @@ jobs:
           RACK_ENV: test
           KNAPSACK_PRO_ENDPOINT: https://api-staging.knapsackpro.com
           KNAPSACK_PRO_TEST_SUITE_TOKEN_RSPEC: 3fa64859337f6e56409d49f865d13fd7
+          EXTRA_TEST_FILES_DELAY: 10
       - image: cimg/postgres:16.1
         environment:
           POSTGRES_DB: rails-app-with-knapsack_pro_test
@@ -116,7 +117,6 @@ jobs:
             # split by test examples ||
             export KNAPSACK_PRO_BRANCH="$CIRCLE_BRANCH--$CIRCLE_BUILD_NUM--regular--split"
             export KNAPSACK_PRO_RSPEC_SPLIT_BY_TEST_EXAMPLES=true
-            export KNAPSACK_PRO_SLOW_TEST_FILE_PATTERN="{spec/timecop_spec.rb}"
             bundle exec rake knapsack_pro:rspec
 
   integration-queue:
@@ -136,6 +136,7 @@ jobs:
           RACK_ENV: test
           KNAPSACK_PRO_ENDPOINT: https://api-staging.knapsackpro.com
           KNAPSACK_PRO_TEST_SUITE_TOKEN_RSPEC: 3fa64859337f6e56409d49f865d13fd7
+          EXTRA_TEST_FILES_DELAY: 10
       - image: cimg/postgres:16.1
         environment:
           POSTGRES_DB: rails-app-with-knapsack_pro_test
@@ -191,7 +192,6 @@ jobs:
             export KNAPSACK_PRO_BRANCH="$CIRCLE_BRANCH--$CIRCLE_BUILD_NUM--queue--split"
             export KNAPSACK_PRO_FIXED_QUEUE_SPLIT=true
             export KNAPSACK_PRO_RSPEC_SPLIT_BY_TEST_EXAMPLES=true
-            export KNAPSACK_PRO_SLOW_TEST_FILE_PATTERN="{spec/timecop_spec.rb}"
             bundle exec rake knapsack_pro:queue:rspec
       - run:
           working_directory: ~/rails-app-with-knapsack_pro
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,45 +1,230 @@
-# Ruby CircleCI 2.0 configuration file
-#
-# Check https://circleci.com/docs/2.0/language-ruby/ for more details
-#
-version: 2
-jobs:
-  build:
-    parallelism: 1
-    docker:
-      # specify the version you desire here
-      # https://circleci.com/developer/images/image/cimg/ruby
-      - image: cimg/ruby:3.2.1
-        environment:
-          CODECLIMATE_REPO_TOKEN: b6626e682a8e97e0c5978febc92c3526792a2d018b41b8e1b52689da37fb7d92
+version: 2.1
 
-    working_directory: ~/knapsack_pro-ruby
-
+commands:
+  setup_knapsack_pro_ruby:
     steps:
       - checkout
-
-      # Download and cache dependencies
       - restore_cache:
           keys:
-          - v2-dependencies-bundler-{{ checksum "knapsack_pro.gemspec" }}
-          # fallback to using the latest cache if no exact match is found
-          - v2-dependencies-bundler-
-
+          - v1-bundler-ruby-{{ checksum "knapsack_pro.gemspec" }}
+          - v1-bundler-ruby-
       - run:
-          name: install ruby dependencies
           command: |
-            bundle install --jobs=4 --retry=3 --path vendor/bundle
-
-      - run: gem install rubocop
-
+            bundle config set --local path './vendor/bundle'
+            bundle install --jobs=4 --retry=3
       - save_cache:
           paths:
             - ./vendor/bundle
-          key: v2-dependencies-bundler-{{ checksum "knapsack_pro.gemspec" }}
+          key: v1-bundler-ruby-{{ checksum "knapsack_pro.gemspec" }}
 
-      # enforce `frozen_string_literal: true` in the gem files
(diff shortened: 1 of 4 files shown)


3v0k4:
@ArturT @shadre how many versions of [Ruby](https://www.ruby-lang.org/en/downloads/releases/) and [RSpec](https://github.com/rspec/rspec-core/blob/main/Changelog.md) do we want to test on CI? I'd say that last 3 of both (that would include Ruby 3.3 as soon as CircleCI publishes the image is a good starting point.

ArturT:
> @ArturT @shadre how many versions of [Ruby](https://www.ruby-lang.org/en/downloads/releases/) and [RSpec](https://github.com/rspec/rspec-core/blob/main/Changelog.md) do we want to test on CI? I'd say that last 3 of both (that would include Ruby 3.3 as soon as CircleCI publishes the image is a good starting point.

**RSpec:**

* We should not test RSpec 2.x because we [dropped support](https://github.com/KnapsackPro/knapsack_pro-ruby/blob/master/CHANGELOG.md#300) for it. 
* The last 3 versions sounds fine. But I'm wondering if we should have some versions fixed. For example 3.10 because we wanted to support some users, fix in [6.0.1](https://github.com/KnapsackPro/knapsack_pro-ruby/blob/master/CHANGELOG.md#601).

If we would keep only last 3 versions and there is a new version of RSpec, should we drop code that supports RSpec 3.10? If so, then we risk that some old users might have again problem to onboard the knapsack_pro gem. I'd just keep 3.10 fixed. 
And keep adding new RSpec versions (latest minor versions).

Example:
* 3.10 (fixed version, we want to test it on CI because our code is specific to this version)
* 3.11.x (not testing anymore on CI)
Last 3 minor versions that we test always:
* 3.12.x
* 3.13.x
* 3.14.x

Other than that we may consider tracking RSpec, Ruby versions in API so that we could better understand the users environment. This way we would know if it makes sense to drop RSpec 3.10 support or not.

**Ruby:**

We can test on CI the last 3 minor versions of Ruby: 3.0.x, 3.1.x, 3.2.x.

As an exception when there is a new Ruby preview/rc release that we want to test for example ruby 3.3.0-rc1 then we can add it to the list. 

Example:
* 3.0.x
* 3.1.x
* 3.2.x
* 3.3.0-rc1

When 3.3 is official then we can get rid of 3.0.x or when it's [not supported](https://endoflife.date/ruby) anymore.

Example:
* 3.1.x
* 3.2.x
* 3.3.x

shadre:
> @ArturT @shadre how many versions of [Ruby](https://www.ruby-lang.org/en/downloads/releases/) and [RSpec](https://github.com/rspec/rspec-core/blob/main/Changelog.md) do we want to test on CI? I'd say that last 3 of both (that would include Ruby 3.3 as soon as CircleCI publishes the image is a good starting point.

@ArturT @3v0k4 I'm leaving the specifics for the both of you to decide on.

My 2 cents is that we don't want to drop testing older versions sooner than the customers realistically bump them. Otherwise, this feature would be providing zero value after some time (as we wouldn't be covering the inevitable bugs that _some_ customers that are slower to update will experience). How that translates to exact delta in versions - I don't know. But I'd like us to keep this point in mind. Another consideration is how much slower it would be for us to test additional versions.

.circleci/config.yml
```diff
@@ -1,45 +1,230 @@
...
+        type: string
+      ruby:
+        type: string
+      rspec:
+        type: string
+    steps:
+      - run:
+          working_directory: << parameters.path >>
+          command: |
+            git clone --depth 1 git@github.com:KnapsackPro/rails-app-with-knapsack_pro.git ./
```
ArturT, line 32: We could use the following. This way you could test na new changes on a feature branch. For example you create a `featureA` branch in the `knapsack_pro-ruby` repository. If you run tests then you will use the `master` branch from `rails-app-with-knapsack_pro` (default).

If you need to add some new tests or changes to `rails-app-with-knapsack_pro` then you don't have to push it to the master. You can create the `featureA` branch in the `rails-app-with-knapsack_pro` with your new tests/changes. Then when you run tests for the `featureA` in `knapsack_pro-ruby` then it would use the `featureA` branch from `rails-app-with-knapsack_pro` to test it on CI.  Once you are happy with new tests/changes in `rails-app-with-knapsack_pro` then you can merge `featureA` into `master`. This way you don't have to polluting the `master` branch with tmp code in `rails-app-with-knapsack_pro` until you are sure this code is fine to add to the master.

```suggestion
            git clone --depth 1 --branch $CIRCLE_BRANCH --single-branch git@github.com:KnapsackPro/rails-app-with-knapsack_pro.git ./ || git clone --depth 1 git@github.com:KnapsackPro/rails-app-with-knapsack_pro.git ./
```

We already do something like that [here](https://github.com/KnapsackPro/rails-app-with-knapsack_pro/blob/675a1188f39cf8a67ffb447a4487db50fcf98619/.circleci/config.yml#L46).

.circleci/config.yml
```diff
@@ -1,45 +1,230 @@
...
+            export KNAPSACK_PRO_ENDPOINT=https://api-fake.knapsackpro.com
+            export KNAPSACK_PRO_MAX_REQUEST_RETRIES=1
+            bundle exec rake knapsack_pro:rspec
+      - run:
+          working_directory: ~/rails-app-with-knapsack_pro
+          command: |
+            # split by test examples ||
+            export KNAPSACK_PRO_BRANCH="$CIRCLE_BRANCH--$CIRCLE_BUILD_NUM--regular--split"
+            export KNAPSACK_PRO_RSPEC_SPLIT_BY_TEST_EXAMPLES=true
+            export KNAPSACK_PRO_SLOW_TEST_FILE_PATTERN="{spec/timecop_spec.rb}"
```
ArturT, line 119:
```suggestion
```

I would add `EXTRA_TEST_FILES_DELAY=10` to global env vars for the docker image.

It would slow down the [`spec/slow_shared_examples_spec.rb`](https://github.com/KnapsackPro/rails-app-with-knapsack_pro/blob/675a1188f39cf8a67ffb447a4487db50fcf98619/spec/slow_shared_examples_spec.rb#L4), which we expect to be split by test examples.

Also I noticed the `spec/foo_spec.rb` is already slow 10s for [the build](https://staging.knapsackpro.com/dashboard/organizations/1/projects/55/test_suites/124/builds/fb7d0c29-aa8f-4b3c-a958-cf567ce1ab37#tab-test-files). So it could be potentially automatically split by examples.

This way we would be closer to what people use as a default. Using `KNAPSACK_PRO_SLOW_TEST_FILE_PATTERN` is not a default approach.
When you use `KNAPSACK_PRO_SLOW_TEST_FILE_PATTERN` then there is no request made to the build_distributions#last API to get the list of tests.

--- 

Or we could leave it as it is if you prefer that. At least we are certain that 1 test file is split by examples (no implicit assumption that we have some slow test file that is supposed to be split).

---

Using `KNAPSACK_PRO_SLOW_TEST_FILE_PATTERN` can be helpful if we would expect a specific path to be split by examples and we test this by checking API response for recorded build subsets (but that's not the scope of this story).

.circleci/config.yml
```diff
@@ -1,45 +1,230 @@
...
+        type: string
+    docker:
+      - image: cimg/ruby:<< parameters.ruby >>-browsers
+        environment:
+          PGHOST: 127.0.0.1
+          PGUSER: rails-app-with-knapsack_pro
+          RAILS_ENV: test
+          RACK_ENV: test
+          KNAPSACK_PRO_ENDPOINT: https://api-staging.knapsackpro.com
+          KNAPSACK_PRO_TEST_SUITE_TOKEN_RSPEC: 3fa64859337f6e56409d49f865d13fd7
```
ArturT, line 78:
```suggestion
          KNAPSACK_PRO_TEST_SUITE_TOKEN_RSPEC: 3fa64859337f6e56409d49f865d13fd7
          EXTRA_TEST_FILES_DELAY: 10
```

ArturT (approved):
I added a few small suggestions.

Issue #242, opened 2024-02-26
 NoMethodError:  undefined method `keys' for nil


Alagaesia93:
Hello!

After bumping version to 7.0 my whole RSpec CI fails with this error
` NoMethodError: undefined method 'keys' for nil`

It must be something connected with dotenv and some env variables, but it started failing only after 7.0 version. I tried to check online, but no luck.

If you need anything else, happy to help!

<details>
  <summary>Full error stack</summary>

```
NoMethodError:
         undefined method `keys' for nil
       # <filtered>/vendor/bundle/ruby/3.3.0/gems/dotenv-3.0.2/lib/dotenv/diff.rb:25:in `added'
       # <filtered>/vendor/bundle/ruby/3.3.0/gems/dotenv-3.0.2/lib/dotenv/diff.rb:47:in `any?'
       # <filtered>/vendor/bundle/ruby/3.3.0/gems/dotenv-3.0.2/lib/dotenv.rb:78:in `restore'
       # <filtered>/vendor/bundle/ruby/3.3.0/gems/dotenv-3.0.2/lib/dotenv/autorestore.rb:9:in `block (2 levels) in <main>'
       # <filtered>/vendor/bundle/ruby/3.3.0/gems/rspec-core-3.12.2/lib/rspec/core/example.rb:457:in `instance_exec'
       # <filtered>/vendor/bundle/ruby/3.3.0/gems/rspec-core-3.12.2/lib/rspec/core/example.rb:457:in `instance_exec'
       # <filtered>/vendor/bundle/ruby/3.3.0/gems/rspec-core-3.12.2/lib/rspec/core/hooks.rb:372:in `run'
       # <filtered>/vendor/bundle/ruby/3.3.0/gems/rspec-core-3.12.2/lib/rspec/core/hooks.rb:529:in `block in run_owned_hooks_for'
       # <filtered>/vendor/bundle/ruby/3.3.0/gems/rspec-core-3.12.2/lib/rspec/core/hooks.rb:528:in `each'
       # <filtered>/vendor/bundle/ruby/3.3.0/gems/rspec-core-3.12.2/lib/rspec/core/hooks.rb:528:in `run_owned_hooks_for'
       # <filtered>/vendor/bundle/ruby/3.3.0/gems/rspec-core-3.12.2/lib/rspec/core/hooks.rb:615:in `block in run_example_hooks_for'
[...]

PR #243, opened 2024-03-07
Require rails_helper before configuring RSpec

Pacyfik:
# Story

https://trello.com/c/iyN1GKYS/320-bug-rspec-knapsackpro-700-fails-with-the-dotenv-gem

## Related

https://github.com/KnapsackPro/knapsack_pro-ruby/issues/242

# Description

Version 7.0.0 introduced some fundamental changes, namely fetching, loading and running batches of specs **after** executing suite hooks, so that such hooks are only ran once, not before every batch. As a result, if `rails_helper` is only required in spec files, which is the RSpec default, instead of e.g. in `.rspec`, then some `before(:suite)` hooks, e.g. defined by gems, are registered after suite hooks had already been executed by the test suite. In comparison, RSpec loads all the spec files **before** executing `before(:suite)` hooks.

# Changes

This PR conditionally adds `--require rails_helper` to cli arguments of `KnapsackPro::Runners::Queue::RSpecRunner` so that the Rails app (and its gems) can be loaded before executing `before(:suite)` hooks. That way all the `before(:suite)` hooks are registered properly and can be executed.

# Checklist reminder

### DIFF
--- a/lib/knapsack_pro/adapters/rspec_adapter.rb
+++ b/lib/knapsack_pro/adapters/rspec_adapter.rb
@@ -52,6 +52,10 @@ module KnapsackPro
         !!parsed_options(cli_args)&.[](:formatters)
       end
 
+      def self.has_require_rails_helper_option?(cli_args)
+        (parsed_options(cli_args)&.[](:requires) || []).include?("rails_helper")
+      end
+
       def self.order_option(cli_args)
         parsed_options(cli_args)&.[](:order)
       end
@@ -74,6 +78,10 @@ module KnapsackPro
       def self.parse_file_path(id)
         # https://github.com/rspec/rspec-core/blob/1eeadce5aa7137ead054783c31ff35cbfe9d07cc/lib/rspec/core/example.rb#L122
         id.match(/\A(.*?)(?:\[([\d\s:,]+)\])?\z/).captures.first
+      end
+
+      def self.rails_helper_exists?(test_dir)
+        File.exist?("#{test_dir}/rails_helper.rb")
       end
 
       # private
--- a/spec/integration/runners/queue/rspec_runner_spec.rb
+++ b/spec/integration/runners/queue/rspec_runner_spec.rb
@@ -372,7 +372,7 @@ describe "#{KnapsackPro::Runners::Queue::RSpecRunner} - Integration tests", :cle
 
       actual = subject
 
-      expect(actual.stdout.scan(/--require rails_helper/).size).to eq 0
+      expect(actual.stdout).to_not include('--require rails_helper')
 
       expect(actual.exit_code).to eq 0
     end
@@ -431,7 +431,7 @@ describe "#{KnapsackPro::Runners::Queue::RSpecRunner} - Integration tests", :cle
 
       actual = subject
 
-      expect(actual.stdout.scan(/--require rails_helper/).size).to eq 3
+      expect(actual.stdout).to include('--require rails_helper')
       expect(actual.stdout.scan(/RSpec_before_suite_hook_from_rails_helper/).size).to eq 1
       expect(actual.stdout.scan(/RSpec_after_suite_hook_from_rails_helper/).size).to eq 1
 
@@ -465,6 +465,7 @@ describe "#{KnapsackPro::Runners::Queue::RSpecRunner} - Integration tests", :cle
       SPEC
 
       spec_c = Spec.new('c_spec.rb', <<~SPEC)
+        require 'rails_helper'
         describe 'C_describe' do
           it 'C1 test example' do
             expect(1).to eq 1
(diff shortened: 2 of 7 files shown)


lib/knapsack_pro/version.rb
```diff
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 module KnapsackPro
-  VERSION = '7.0.0'
+  VERSION = '7.0.1'
```
ArturT, line 4: No need to update the gem version inside of the PR.

We increase the gem version after the PR is merged. We create a commit that updates the version on the master branch, and then we create a git tag for that commit.

Sorry, the docs are not clear enough here:
https://github.com/KnapsackPro/knapsack_pro-ruby/blob/master/README.md#publishing
* CHANGELOG.md should be updated in PR
* lib/knapsack_pro/version.rb should be updated after PR is merged, and then you follow the next steps from the readme

Some benefits of this approach:
* while you are testing the gem locally you won't use a version number that has not been released yet. You won't pollute the local API DB or staging/production DB if you hit our API with this unreleased version.
*  when we create a git tag on master, it's for a bumped version commit, so the git history is clear

Pacyfik: Ah, of course, it makes a lot of sense. I hadn't thought of the implications, e.g. sending a new version to the api when testing.

ArturT: Would you help to update the readme about publishing to make it more clear?

Pacyfik: Done [here](https://github.com/KnapsackPro/knapsack_pro-ruby/pull/244).

ArturT (commented):
Good direction. I added some comments. 

We will need integration tests in `spec/integration/runners/queue/rspec_runner_spec.rb`. We want to cover case when the `rails_helper.rb` file exists and it has RSpec before(:suite), after(:suite) hooks defined and only a spec file in 2nd batch of tests requires `rails_helper`.

spec/knapsack_pro/adapters/rspec_adapter_spec.rb
```diff
@@ -187,6 +187,54 @@
     end
   end
 
+  describe '.has_require_rails_helper_option?' do
+    subject { described_class.has_require_rails_helper_option?(cli_args) }
+
+    context 'when require option is provided as -r' do
```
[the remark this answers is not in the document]
Pacyfik: I think it's fine as it is, to be honest. And it follows all the other examples.

lib/knapsack_pro/adapters/rspec_adapter.rb
```diff
@@ -76,6 +80,10 @@ def self.parse_file_path(id)
         id.match(/\A(.*?)(?:\[([\d\s:,]+)\])?\z/).captures.first
       end
 
+      def self.rails_helper_exists?
+        File.exist?('spec/rails_helper.rb')
```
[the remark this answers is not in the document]
Pacyfik: Great call. And it made me rethink my approach a bit.

spec/integration/runners/queue/rspec_runner_spec.rb
```diff
@@ -327,6 +332,173 @@ def log_command_result(stdout, stderr, status)
...
+      SPEC
+
+      generate_specs(spec_helper, rspec_options, [
+        [spec_a, spec_b],
+        [spec_c],
+      ])
+
+      actual = subject
+
+      expect(actual.stdout.scan(/--require rails_helper/).size).to eq 0
```
ArturT, line 375:
```suggestion
      expect(actual.stdout).to_not include('--require rails_helper')
```

I think this would be simpler and more consistent with other examples. What do you think?

spec/integration/runners/queue/rspec_runner_spec.rb
```diff
@@ -327,6 +332,173 @@ def log_command_result(stdout, stderr, status)
...
+          puts 'RSpec_after_suite_hook_from_rails_helper'
+        end
+      end
+      SPEC
+
+      create_rails_helper_file(rails_helper)
+
+      actual = subject
+
+      expect(actual.stdout.scan(/--require rails_helper/).size).to eq 3
```
ArturT, line 434:
```suggestion
      expect(actual.stdout).to include('--require rails_helper')
```

I think it's enough to check the presence once.

Counting `--require rails_helper` makes the test more dependent on another feature. The feature which involves printing an RSpec command that can be easily copied & pasted to reproduce a batch of tests.

spec/integration/runners/queue/rspec_runner_spec.rb
```diff
@@ -327,6 +332,173 @@ def log_command_result(stdout, stderr, status)
...
+        require 'rails_helper'
+        describe 'B_describe' do
+          it 'B1 test example' do
+            expect(1).to eq 1
+          end
+        end
+      SPEC
+
+      spec_c = Spec.new('c_spec.rb', <<~SPEC)
+        describe 'C_describe' do
```
ArturT, line 468:
```suggestion
        require 'rails_helper'
        describe 'C_describe' do
```

Perhaps we could require rails_helper in the second batch as well, just to ensure that the number of batches doesn't affect how many times the RSpec hooks are called.

ArturT (approved):
Good work!

PR #249, opened 2024-04-30
[RSpec] [Queue Mode] Log error message and backtrace when unexpected failure raised in process

tubaxenor:
# Description

Currently when unexpected error raised, we only get the class name from `exception.inspect`, this PR adds details to the logger to be able to trace or debug the cause.

# Changes

Add more logs when unexpected error raised in runner process.

# Checklist reminder


ArturT:
Thank you for the contribution. We released the knapsack_pro v7.1.0 gem version.

PR #266, opened 2024-08-06
Log threads when the OS signal is received to simplify debugging, especially when a CI node hangs.

ArturT:
# Story

https://trello.com/c/oQCHRpec

## Related

https://docs.knapsackpro.com/ruby/troubleshooting/#why-does-the-ci-node-hang-backtrace-debugging

# Description

Log threads when the OS signal is received to simplify debugging, especially when a CI node hangs.

# Checklist reminder

### DIFF
--- a/lib/knapsack_pro/runners/queue/base_runner.rb
+++ b/lib/knapsack_pro/runners/queue/base_runner.rb
@@ -69,7 +69,6 @@ module KnapsackPro
 
           puts
           puts '=' * 80
-          puts "Process pid: #{Process.pid}"
           puts "Start logging #{threads.count} detected threads."
           puts 'See the following backtrace and the line of code that is stuck if your CI node hangs.'
 
--- a/lib/knapsack_pro/runners/queue/base_runner.rb
+++ b/lib/knapsack_pro/runners/queue/base_runner.rb
@@ -75,7 +75,7 @@ module KnapsackPro
           threads.each do |thread|
             puts
             if thread == Thread.main
-              puts "Main thread and its backtrace:"
+              puts "Main thread backtrace:"
             else
               puts "Non-main thread inspect: #{thread.inspect}"
               puts "Non-main thread backtrace:"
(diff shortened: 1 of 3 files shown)


lib/knapsack_pro/runners/queue/base_runner.rb
```diff
@@ -59,9 +59,37 @@ def trap_signals
...
+            puts
+            if thr == Thread.main
+              puts "Main thread and its backtrace:"
+            else
+              puts "Non-main thread inspect: #{thr.inspect}"
+              puts "Non-main thread backtrace:"
+            end
+            puts thr.backtrace.join("\n")
+            puts
+          end
```
3v0k4, line 85: suggestion

Would you be willing not to abbreviate? I'd prefer to avoid abbreviations as much as possible.

```suggestion
          threads.each do |thread|
            puts
            if thread == Thread.main
              puts "Main thread and its backtrace:"
            else
              puts "Non-main thread inspect: #{thread.inspect}"
              puts "Non-main thread backtrace:"
            end
            puts thread.backtrace.join("\n")
            puts
          end
```

lib/knapsack_pro/runners/queue/base_runner.rb
```diff
@@ -59,9 +59,37 @@ def trap_signals
...
+
+          puts
+          puts '=' * 80
+          puts "Start logging #{threads.count} detected threads."
+          puts 'Use the following backtrace to find the line of code that got stuck if the CI node hung and terminated your tests.'
+
+          threads.each do |thr|
+            puts
+            if thr == Thread.main
+              puts "Main thread and its backtrace:"
```
3v0k4, line 78: recommendation

Maybe we can use the same string as the one for the non-main thread?

```suggestion
              puts "Main thread backtrace:"
```

lib/knapsack_pro/runners/queue/base_runner.rb
```diff
@@ -59,9 +59,37 @@ def trap_signals
...
           end
         end
+
+        def log_threads
+          threads = Thread.list
+
+          puts
+          puts '=' * 80
+          puts "Start logging #{threads.count} detected threads."
+          puts 'Use the following backtrace to find the line of code that got stuck if the CI node hung and terminated your tests.'
```
3v0k4, line 73:
```suggestion
          puts 'Use the following backtrace(s) to find the line of code that got stuck if the CI node hung and terminated your tests.'
```

3v0k4 (commented):
🙌

PR #276, opened 2024-10-22
slow test split patch

devinburnette:
# Story

TODO: link to the internal story

## Related

TODO: links to related PRs or issues

# Description

I tried using `KNAPSACK_PRO_BRANCH` to uniquely set branch names so that a KNAPSACK_TOKEN can be shared across multiple projects. I noticed an issue where because of this, `BuildDistributions.last` returns every recorded test, and the cumulative time execution is too great for the `KNAPSACK_PRO_RSPEC_SPLIT_BY_TEST_EXAMPLES` feature to continue to work correctly. 

# Changes

I used this change to monkey patch that functionality to only consider the cumulative time execution of only the tests on disk instead of the entire recorded set of tests returned by the build distributions API. Not married to this approach, but if it makes it easier to reason about a more permanent fix, please consider this as a starting point. Feel free to edit and build from this.

# Checklist reminder

### DIFF
--- a/lib/knapsack_pro/adapters/base_adapter.rb
+++ b/lib/knapsack_pro/adapters/base_adapter.rb
@@ -22,14 +22,12 @@ module KnapsackPro
       def self.slow_test_file?(adapter_class, test_file_path)
         @slow_test_file_paths ||=
           begin
-            slow_test_files =
-              if KnapsackPro::Config::Env.slow_test_file_pattern
-                KnapsackPro::TestFileFinder.slow_test_files_by_pattern(adapter_class)
-              else
-                # get slow test files from JSON file based on data from API
-                KnapsackPro::SlowTestFileDeterminer.read_from_json_report
-              end
-            KnapsackPro::TestFilePresenter.paths(slow_test_files)
+            slow_test_files = []
+            if KnapsackPro::Config::Env.slow_test_file_pattern
+              slow_test_files << KnapsackPro::TestFileFinder.slow_test_files_by_pattern(adapter_class)
+            end
+            slow_test_files << KnapsackPro::SlowTestFileDeterminer.read_from_json_report
+            KnapsackPro::TestFilePresenter.paths(slow_test_files.flatten.uniq)
           end
         clean_path = KnapsackPro::TestFileCleaner.clean(test_file_path)
         @slow_test_file_paths.include?(clean_path)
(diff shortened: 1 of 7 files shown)


devinburnette:
@ArturT a couple additional ideas that I think my team finds helpful with regard to this slow test splitting feature.
1. make the slow test time threshold per ci node configurable, this will allow teams to find a sweet spot and even on a per project basis if they want to.
2. allowing teams to pass in slow test patterns AND have it automatically detect as well, right now it only supports either or, but not both. this is helpful in dynamic parallelism scenarios where changing the parallelism affects the time threshold calculation and as a result may miss tests that should always be split.

ArturT:
I think doing 1 and lowering the threshold would make 2 not needed then. 

BTW are you sure 1 is needed? If you lower the threshold of tests then more tests are split by examples and this also has impact on speed and memory usage by RSpec. 70% threshold is a sweet spot right now.

ArturT:
@devinburnette Thank you for your contribution and for suggesting ideas. 

Slow tests are dynamically determined to be well-balanced across CI nodes. If you want to split more tests by test examples, just run more parallel nodes. It will happen automatically.

I'm closing this PR in favor of this one:
https://github.com/KnapsackPro/knapsack_pro-ruby/pull/277

We released the knapsack_pro 7.10.0 version. Please update it in your project. Thanks.

devinburnette:
Thanks!

> Slow tests are dynamically determined to be well-balanced across CI nodes. If you want to split more tests by test examples, just run more parallel nodes. It will happen automatically.

@ArturT yes, but the goal is for the number of parallel nodes to be dynamic, in order to support that, I think that means we'll at least need number 2 above. There are certain test files that we want to ensure always get split. But we want to also keep the automatic calculation as well. Can we at least support that? I think it's valuable for knapsack to take a merged set of these slow test files. Both automatic and manually supplied.

devinburnette:
The idea is this, some test runs will purposely skip a lot of tests, and therefore we don't want to run the full set of parallel CI nodes only a few. And so that will impact the calculation of slow test files and some of the longest running test files will not be ran as a result. So we need a way of tell it to always run those few test files in those cases.

ArturT:
@devinburnette In an ideal scenario, you run tests without needing to manually define slow tests for splitting by test examples - it happens automatically.

I believe the situation you’re describing should already be handled by the API, and slow test files should be properly detected. If this isn’t happening with knapsack_pro 7.10.0, please provide links to a few CI builds where slow test files should have been split by test examples but were not. Also, include the text output from CI (ensuring you use debug log level). You can send it to support@knapsackpro.com.

I’d be happy to investigate and determine if any improvements are needed on the API or gem side. Thank you!

PR #282, opened 2024-10-31
Add `KNAPSACK_PRO_SLOW_TEST_FILE_THRESHOLD` to improve the RSpec split by examples feature with many skipped tests 

ArturT:
# Story

TODO: [link to the internal story](https://trello.com/c/f8uD3QC0)

## Related

* https://github.com/KnapsackPro/knapsack_pro-ruby/pull/281

# Description

Relate to the RSpec split by examples feature:
https://docs.knapsackpro.com/ruby/split-by-test-examples/

If a user skips some test files, for example, using the RSpec `--tag` option, then RSpec might still load a file but not run it or only run some of the test examples from it.

If a lot of test files are skipped, then the actual average CI node execution time would be much smaller than the estimated time (Knapsack Pro does not know how many test files are filtered out by `--tag`).  This can lead to not determining enough test files as slow, which examples should split. As a result, CI build is unbalanced.

To solve that problem, we added a new environment variable (`KNAPSACK_PRO_SLOW_TEST_FILE_THRESHOLD`). You define a threshold (a number of seconds) for slow test files. Any test file above the threshold would be split by test examples.
For example, you expect your CI node to run no longer than 3 minutes, then set `KNAPSACK_PRO_SLOW_TEST_FILE_THRESHOLD=180` (180 seconds == 3 minutes)

## changes

* Add the `KNAPSACK_PRO_SLOW_TEST_FILE_THRESHOLD` option
* do not split test files by test cases if you run tests on a single CI node to run Knapsack Pro faster

# Checklist reminder

### DIFF
--- a/lib/knapsack_pro/config/env.rb
+++ b/lib/knapsack_pro/config/env.rb
@@ -201,12 +201,11 @@ module KnapsackPro
         end
 
         def slow_test_file_threshold
-          value = ENV.fetch('KNAPSACK_PRO_SLOW_TEST_FILE_THRESHOLD', nil)
-          value.to_f unless value.nil?
+          ENV.fetch('KNAPSACK_PRO_SLOW_TEST_FILE_THRESHOLD', nil)&.to_f
         end
 
         def slow_test_file_threshold?
-          !slow_test_file_threshold.nil?
+          !!slow_test_file_threshold
         end
 
         def test_suite_token
--- a/lib/knapsack_pro/config/env.rb
+++ b/lib/knapsack_pro/config/env.rb
@@ -200,6 +200,15 @@ module KnapsackPro
           ENV.fetch('KNAPSACK_PRO_RSPEC_TEST_EXAMPLE_DETECTOR_PREFIX', 'bundle exec')
         end
 
+        def slow_test_file_threshold
+          value = ENV.fetch('KNAPSACK_PRO_SLOW_TEST_FILE_THRESHOLD', nil)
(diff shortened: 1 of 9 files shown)


lib/knapsack_pro/config/env.rb
```diff
@@ -200,6 +200,15 @@ def rspec_test_example_detector_prefix
           ENV.fetch('KNAPSACK_PRO_RSPEC_TEST_EXAMPLE_DETECTOR_PREFIX', 'bundle exec')
         end
 
+        def slow_test_file_threshold
+          value = ENV.fetch('KNAPSACK_PRO_SLOW_TEST_FILE_THRESHOLD', nil)
+          value.to_f unless value.nil?
```
shadre, line 205: It seems very unobvious to me that this line's return value would be `nil` if `value` is `nil`. Yet, we are relying on the strict return value of the method on L209.

I propose to make one (or both) of the below changes:

L205:
```ruby
value&.to_f
```
L208:
```ruby
!!slow_test_file_threshold
```

lib/knapsack_pro/slow_test_file_determiner.rb
```diff
@@ -11,7 +11,11 @@ def self.call(test_files)
 
       test_files.select do |test_file|
         time_execution = test_file.fetch('time_execution')
```
shadre, line 13: I would change the local var to `execution_time`

shadre (commented):
Added suggestions

Issue #285, opened 2024-11-21
What test was running during a signal trap?


devinburnette:
Right now we get the backtrace of all the threads when trapping a signal from a timeout for example, but it's still difficult to pin point which test the calls are coming from..

Is there a way to output the test or tests that were running at the time of the signal trap? We get the "retry the previous batch of tests" in the INFO logs, but not for the current batch of tests..

ArturT:
The spec file could be in the main thread backtrace as shown here:
https://docs.knapsackpro.com/ruby/troubleshooting/#why-does-the-ci-node-hang-backtrace-debugging

If not, then something else could freeze, like factories (FactoryBot) etc.

You may also see the name of the test case in the output right before the thread backtraces if you use rspec `--format documentation`. It's easier to catch the problematic test case then.


Feel free to share the full txt output from a node that hangs to the support email https://knapsackpro.com/contact
We can analyze it. Perhaps there is a better way to narrow down the root issue or automate it somehow so the output is more readable.

devinburnette:
It may be helpful to include the seed in the debug logs BEFORE a batch of tests are executed, in the case of a signal trap due to timeout, we never get to see the seed and it makes it difficult to try to reproduce the issue locally.

PR #286, opened 2024-12-03
Fix handling signals for non-RSpec test runners

ArturT:
# Story

N/A

# Description

Fix handling signals for non-RSpec test runners.

# Changes

* fix(bug): we should set RSpec wants to quite value in the RSpec only. Remove it from the queue base runner which breaks handling signals for non RSpec test runners.

# Checklist reminder

### DIFF
--- a/lib/knapsack_pro/runners/queue/rspec_runner.rb
+++ b/lib/knapsack_pro/runners/queue/rspec_runner.rb
@@ -117,6 +117,10 @@ module KnapsackPro
 
         private
 
+        def post_trap_signals
+          RSpec.world.wants_to_quit = true
+        end
+
         def pre_run_setup
           ENV['KNAPSACK_PRO_QUEUE_RECORDING_ENABLED'] = 'true'
           ENV['KNAPSACK_PRO_QUEUE_ID'] = KnapsackPro::Config::EnvGenerator.set_queue_id
--- a/lib/knapsack_pro/runners/queue/base_runner.rb
+++ b/lib/knapsack_pro/runners/queue/base_runner.rb
@@ -59,10 +59,13 @@ module KnapsackPro
             Signal.trap(signal) {
               puts "#{signal} signal has been received. Terminating Knapsack Pro..."
               @@terminate_process = true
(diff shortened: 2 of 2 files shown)


lib/knapsack_pro/runners/queue/rspec_runner.rb
```diff
@@ -117,6 +117,10 @@ def log_fail_fast_limit_met
 
         private
 
+        def post_trap_signals
+          RSpec.world.wants_to_quit = true
```
ArturT, line 121: @3v0k4 I'm thinking about adding the following seed here so that we could help with [the issue mentioned here](https://github.com/KnapsackPro/knapsack_pro-ruby/issues/285#issuecomment-2514693380). What do you think?

```suggestion
          puts "Randomized with seed #{@rspec_runner.knapsack__seed.value}"
          RSpec.world.wants_to_quit = true
```

The output would look like this when a signal is received:

```
^CINT signal has been received. Terminating Knapsack Pro...
Randomized with seed 44874

================================================================================
Start logging 3 detected threads.
Use the following backtrace(s) to find the line of code that got stuck if the CI node hung and terminated your tests.
How to read the backtrace: https://knapsackpro.com/perma/ruby/backtrace-debugging

Main thread backtrace:
...
```

3v0k4: Doesn't RSpec already log the seed?

3v0k4: Also, I'd expect this to happen in a separate PR.

3v0k4 (approved):
I'm wondering if we should add an [integration test](https://github.com/KnapsackPro/knapsack_pro-ruby/tree/master/spec/integration/runners/queue) that exercises the `trap`. I trust your judgement on this.

Do we have a similar problem in Regular mode?

Feel free to merge & release when convenient.

ArturT:
> Do we have a similar problem in Regular mode?

No. Test runners handle signals on their own.

PR #287, opened 2025-01-20
Improve debugging for hanging CI node. Show hanging spec files in the RSpec output and a command to reproduce the current batch of tests 

ArturT:
# Story

https://trello.com/c/xvarNSRi

## Related

Issue:

* https://github.com/KnapsackPro/knapsack_pro-ruby/issues/285

Related PR:

* https://github.com/KnapsackPro/docs.knapsackpro.com/pull/229

# Description

Improve debugging for hanging CI node. 

* Show hanging spec files in the RSpec output.
* Show a command to reproduce the current batch of tests. It includes seed if the RSpec tests were running in random order.

# Checklist reminder

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### 7.14.0
 
-* Improve debugging for hanging CI node. Show hanging spec files in the RSpec output and a command to reproduce the current batch of tests.
+* Improve debugging for hanging CI nodes: show hanging spec files in the RSpec output and a command to reproduce the current batch of tests.
 
     https://github.com/KnapsackPro/knapsack_pro-ruby/pull/287
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+### 7.14.0
+
+* Improve debugging for hanging CI node. Show hanging spec files in the RSpec output and a command to reproduce the current batch of tests.
+
+    https://github.com/KnapsackPro/knapsack_pro-ruby/pull/287
+
+https://github.com/KnapsackPro/knapsack_pro-ruby/compare/v7.13.1...v7.14.0
+
 ### 7.13.1
 
 * Fix handling signals for non-RSpec test runners
--- a/lib/knapsack_pro/runners/queue/base_runner.rb
+++ b/lib/knapsack_pro/runners/queue/base_runner.rb
@@ -77,6 +77,8 @@ module KnapsackPro
           puts 'Use the following backtrace(s) to find the line of code that got stuck if the CI node hung and terminated your tests.'
           puts 'How to read the backtrace: https://knapsackpro.com/perma/ruby/backtrace-debugging'
 
+          post_log_threads(threads)
+
           threads.each do |thread|
             puts
             if thread == Thread.main
@@ -94,6 +96,9 @@ module KnapsackPro
           puts '=' * 80
 
           $stdout.flush
+        end
+
+        def post_log_threads
         end
       end
     end
--- a/lib/knapsack_pro/runners/queue/base_runner.rb
+++ b/lib/knapsack_pro/runners/queue/base_runner.rb
@@ -98,7 +98,8 @@ module KnapsackPro
           $stdout.flush
         end
 
-        def post_log_threads
+        def post_log_threads(threads)
+          # implement in a child class if you need to log more info
         end
       end
     end
--- a/lib/knapsack_pro/runners/queue/base_runner.rb
+++ b/lib/knapsack_pro/runners/queue/base_runner.rb
@@ -77,7 +77,7 @@ module KnapsackPro
           puts 'Use the following backtrace(s) to find the line of code that got stuck if the CI node hung and terminated your tests.'
           puts 'How to read the backtrace: https://knapsackpro.com/perma/ruby/backtrace-debugging'
 
-          post_log_threads(threads)
+          log_current_tests(threads)
 
           threads.each do |thread|
             puts
@@ -98,7 +98,7 @@ module KnapsackPro
           $stdout.flush
         end
 
-        def post_log_threads(threads)
+        def log_current_tests(threads)
           # implement in a child class if you need to log more info
         end
       end
--- a/lib/knapsack_pro/runners/queue/rspec_runner.rb
+++ b/lib/knapsack_pro/runners/queue/rspec_runner.rb
@@ -120,7 +120,7 @@ module KnapsackPro
         def post_trap_signals
           RSpec.world.wants_to_quit = true
 
-          print_current_batch_rspec_command
+          log_current_batch_rspec_command
         end
 
         def log_current_tests(threads)
@@ -186,7 +186,7 @@ module KnapsackPro
           test_file_paths
         end
 
-        def print_current_batch_rspec_command
+        def log_current_batch_rspec_command
           test_file_paths = @queue.current_batch&.test_file_paths
           return unless test_file_paths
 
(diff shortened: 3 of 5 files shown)


spec/integration/runners/queue/rspec_runner_spec.rb
```diff
@@ -1403,12 +1403,17 @@ def when_first_matching_example_defined(type:)
...
 
+      expect(actual.stdout).to include('To retry the last batch of tests fetched from the Queue API, please run the following command on your machine:')
+      expect(actual.stdout).to include('bundle exec rspec --format documentation --default-path spec_integration "spec_integration/b_spec.rb" "spec_integration/c_spec.rb"')
 
       expect(actual.stdout).to include('Use the following backtrace(s) to find the line of code that got stuck if the CI node hung and terminated your tests.')
+      expect(actual.stdout).to include('Hanging specs in the main thread:')
+      expect(actual.stdout).to include('Hanging specs in non-main thread:')
       expect(actual.stdout).to include('Main thread backtrace:')
-      expect(actual.stdout).to match(/spec_integration\/b_spec\.rb:7:in .*kill/)
+      expect(actual.stdout.scan(/spec_integration\/b_spec\.rb:7:in .*kill/).size).to eq 2
```
3v0k4, line 1413: recommendation

This reads a bit better:
```suggestion
      expect(actual.stdout).to match(/spec_integration\/b_spec\.rb:7:in .*kill/).twice
```

spec/integration/runners/queue/rspec_runner_spec.rb
```diff
@@ -1403,12 +1403,17 @@ def when_first_matching_example_defined(type:)
...
+      expect(actual.stdout).to include('Hanging specs in the main thread:')
       expect(actual.stdout).to include('Non-main thread backtrace:')
-      expect(actual.stdout).to match(/spec_integration\/a_spec\.rb:6:in .*sleep/)
+      expect(actual.stdout.scan(/spec_integration\/a_spec\.rb:6:in .*sleep/).size).to eq 2
```
3v0k4, line 1416: recommendation

This reads a bit better:
```suggestion
      expect(actual.stdout).to match(/spec_integration\/a_spec\.rb:6:in .*sleep/).twice
```

lib/knapsack_pro/runners/queue/rspec_runner.rb
```diff
@@ -165,6 +186,21 @@ def pull_tests_from_queue(can_initialize_queue: false)
           test_file_paths
         end
 
+        def print_current_rspec_batch_command
```
3v0k4, line 189: suggestion

I'd keep current & batch and rspec & command together:
```suggestion
        def print_current_batch_rspec_command
```

lib/knapsack_pro/runners/queue/rspec_runner.rb
```diff
@@ -165,6 +186,21 @@ def pull_tests_from_queue(can_initialize_queue: false)
...
+          test_file_paths = @queue.current_batch&.test_file_paths
```
3v0k4, line 190: When is the `current_batch` nil? Is it when the trap happened before/in between/after a batch?

lib/knapsack_pro/runners/queue/rspec_runner.rb
```diff
@@ -165,6 +186,21 @@ def pull_tests_from_queue(can_initialize_queue: false)
...
+          return unless test_file_paths
+
+          puts
+          puts '=' * 80
+
+          order_option = @adapter_class.order_option(@cli_args)
+          printable_args = @rspec_pure.args_with_seed_option_added_when_viable(order_option, @rspec_runner.knapsack__seed, @cli_args)
+          messages = @rspec_pure.rspec_command(printable_args, test_file_paths, :batch_finished)
```
3v0k4, line 198: It's not really true that the `:batch_finished`, but I guess it does not matter, right?

lib/knapsack_pro/runners/queue/rspec_runner.rb
```diff
@@ -119,6 +119,27 @@ def log_fail_fast_limit_met
...
+
+          print_current_rspec_batch_command
+        end
+
+        def post_log_threads(threads)
+          threads.each do |thread|
+            next unless thread.backtrace
+
+            spec_file_lines = thread.backtrace.select { |line| line.include?('_spec.rb') }
+            next if spec_file_lines.empty?
```
3v0k4, line 131: suggestion

`empty?` feels to broad, what about `.nil?` or `unless spec_file_lines`?

lib/knapsack_pro/runners/queue/rspec_runner.rb
```diff
@@ -119,6 +119,27 @@ def log_fail_fast_limit_met
...
+
+            puts
+            if thread == Thread.main
+              puts "Hanging specs in the main thread:"
+            else
+              puts "Non-main thread inspect: #{thread.inspect}"
+              puts "Hanging specs in non-main thread:"
```
3v0k4, line 138: plea

Hanging means that the code is actually stuck. But here it's about what's running when the method is executed. What about "Running specs"?

lib/knapsack_pro/runners/queue/base_runner.rb
```diff
@@ -77,6 +77,8 @@ def log_threads
           puts 'Use the following backtrace(s) to find the line of code that got stuck if the CI node hung and terminated your tests.'
           puts 'How to read the backtrace: https://knapsackpro.com/perma/ruby/backtrace-debugging'
 
+          post_log_threads(threads)
```
3v0k4, line 80: plea

"Post" does not work because this happens in the middle of `log_threads`. What about `log_current_tests`?

Also, we are using both log and print, maybe we should converge into one?

lib/knapsack_pro/runners/queue/rspec_runner.rb
ArturT: Yes. Trap could happen before tests are pulled from the Queue API.

lib/knapsack_pro/runners/queue/rspec_runner.rb
ArturT: Yes.

It would print  the `To retry the last batch of tests fetched from the Queue API, please run the following command on your machine:` message.

lib/knapsack_pro/runners/queue/rspec_runner.rb
ArturT: `select` returns `[]` so empty? is fine.

spec/integration/runners/queue/rspec_runner_spec.rb
ArturT: It does not work.

```
  Failure/Error: expect(actual.stdout).to match(/spec_integration\/b_spec\.rb:7:in .*kill/).twice

     NoMethodError:
       undefined method `twice' for an instance of RSpec::Matchers::BuiltIn::Match
```

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Changelog
 
+### 7.14.0
+
+* Improve debugging for hanging CI node. Show hanging spec files in the RSpec output and a command to reproduce the current batch of tests.
```
3v0k4, line 5:
```suggestion
* Improve debugging for hanging CI nodes: show hanging spec files in the RSpec output and a command to reproduce the current batch of tests.
```

lib/knapsack_pro/runners/queue/base_runner.rb
3v0k4: There's an inconsistency: `print_current_batch_rspec_command` uses `puts`, `log_current_tests` uses `puts`.

So they both use `puts` but why is one called "print" and the other "log"?

3v0k4 (approved):

ArturT:
> There's an inconsistency: print_current_batch_rspec_command uses puts, log_current_tests uses puts.

> So they both use puts but why is one called "print" and the other "log"?

Ok. I'll do log_current_batch_rspec_command. log is a higher abstraction than print.

PR #288, opened 2025-02-17
Enable Split By Test Example by default in RSpec

3v0k4:
- **chore: sort deps alphabetically**
- **chore: improve code comments**
- **chore: enable split by test examples by default**

# Story

TODO: link to the internal story

## Related

TODO: links to related PRs or issues

# Description

TODO

# Changes

TODO: changes introduced by this PR

# Checklist reminder


lib/knapsack_pro/test_case_detectors/rspec_test_example_detector.rb
```diff
@@ -3,7 +3,7 @@
 module KnapsackPro
   module TestCaseDetectors
     class RSpecTestExampleDetector
-      def generate_json_report
+      def generate_json_report(args)
```
3v0k4, line 6: plea

```suggestion
      def generate_json_report(rspec_args)
```

lib/knapsack_pro/test_case_detectors/rspec_test_example_detector.rb
```diff
@@ -25,7 +25,10 @@ def generate_json_report
           return
         end
 
-        cli_args = cli_format + [
+        raw_cli_args = (args || '').split
+        cli_args_without_formatters = KnapsackPro::Adapters::RSpecAdapter.remove_formatters(raw_cli_args)
```
3v0k4, line 29: suggestion

```suggestion
        args = (rspec_args || '').split
        cli_args_without_formatters = KnapsackPro::Adapters::RSpecAdapter.remove_formatters(args)
```

spec/integration/runners/queue/rspec_runner_spec.rb
```diff
@@ -2060,29 +2057,26 @@ def when_first_matching_example_defined(type:)
 
   context 'when the RSpec split by test examples is enabled AND --tag is set' do
     before do
-      ENV['KNAPSACK_PRO_RSPEC_SPLIT_BY_TEST_EXAMPLES'] = 'true'
-
-      # remember to stub Queue API batches to include test examples (example: a_spec.rb[1:1])
+      # remember to stub Queue API batches to include test examples (example: a_spec.rb[1:2)
       # for the following slow test files
```
3v0k4, line 2061: plea

```suggestion
      # Remember to stub the Queue API batches to include test examples (example: a_spec.rb[1:2])
      # for the following slow test files.
```

spec/integration/runners/queue/rspec_runner_spec.rb
```diff
@@ -2114,43 +2108,39 @@ def when_first_matching_example_defined(type:)
         [spec_a, spec_b, spec_c]
       ])
       stub_test_cases_for_slow_test_files([
-        "#{spec_a.path}[1:1]",
-        "#{spec_a.path}[1:2]",
+        "#{spec_a.path}[1:2]", # only this test example has is tagged with mytag
```
3v0k4, line 2111: plea

```suggestion
        "#{spec_a.path}[1:2]", # only this test example is tagged
```

3v0k4: Are you passing the id path on purpose or did you want the file path?

spec/knapsack_pro/test_case_detectors/rspec_test_example_detector_spec.rb
```diff
@@ -4,7 +4,7 @@
   let(:rspec_test_example_detector) { described_class.new }
 
   describe '#generate_json_report' do
-    subject { rspec_test_example_detector.generate_json_report }
+    subject { rspec_test_example_detector.generate_json_report(args) }
```
3v0k4, line 7:
```suggestion
    subject { rspec_test_example_detector.generate_json_report(rspec_args) }
```

spec/knapsack_pro/test_case_detectors/rspec_test_example_detector_spec.rb
```diff
@@ -45,7 +45,7 @@
           expect(KnapsackPro::TestFilePattern).to receive(:test_dir).with(KnapsackPro::Adapters::RSpecAdapter).and_return(test_dir)
 
           options = double
-          expect(RSpec::Core::ConfigurationOptions).to receive(:new).with([
+          expect(RSpec::Core::ConfigurationOptions).to receive(:new).with(expected_raw_cli_args + [
```
3v0k4, line 48:
```suggestion
          expect(RSpec::Core::ConfigurationOptions).to receive(:new).with(expected_args + [
```

lib/knapsack_pro/runners/queue/rspec_runner.rb
```diff
@@ -21,6 +21,7 @@ def self.run(args, stream_error = $stderr, stream_out = $stdout)
         end
 
         def initialize(adapter_class, rspec_pure, args, stream_error, stream_out)
+          ENV['KNAPSACK_PRO_RSPEC_OPTIONS'] = args
```
3v0k4, line 24: Have you considered doing it (and for Regular) in the Rake task?

I don't have a strong opinion either way, so if this is simpler (eg, tests), I'm fine with it.

3v0k4: Sometimes we are calling it args, sometimes options. I think we should stick with options for this one (vs positional arguments.)

lib/tasks/rspec.rake
```diff
@@ -13,6 +13,6 @@ namespace :knapsack_pro do
     ENV.delete('SPEC_OPTS')
 
     detector = KnapsackPro::TestCaseDetectors::RSpecTestExampleDetector.new
-    detector.generate_json_report
+    detector.generate_json_report(ENV['KNAPSACK_PRO_RSPEC_OPTIONS'])
```
3v0k4, line 16: Should we raise if the ENV is not set?

3v0k4 (commented):
There's an important thing missing: the formatters are removed from the rspec options but *not* the ones coming from configuration files (eg, `.rspec`).

I think it would be simpler to allow formatters without manually parsing them out (fragile code). This would make all the formatters w/o `--out` print to stdout, but we could either:
- Silence stdout but not stderr `> /dev/null`; we need to ensure we are not silencing anything that is needed for debug (in theory, errors should be printed to stderr, but most people throw everything to stdout)
- Append `--out /dev/null` (or something like it) to all the existing formatters; we need to test what happens with `-f json -o /dev/null -f json -o /path/to/report`
- Other ideas?

---

Are we also supporting `--example` and `--example-matches` with this PR?

lib/tasks/rspec.rake
ArturT: No. It's fine if people do not pass any args to the Knapsack Pro command like:

```
bundle exec rake knapsack_pro:queue:rspec
```

ENV['KNAPSACK_PRO_RSPEC_OPTIONS'] is nil then, and this is fine.

lib/knapsack_pro/runners/queue/rspec_runner.rb
ArturT: > Have you considered doing it (and for Regular) in the Rake task?

No.

> I don't have a strong opinion either way, so if this is simpler (eg, tests), I'm fine with it.

It's simpler to have it in runner classes like `lib/knapsack_pro/runners/queue/rspec_runner.rb` for testing purposes.

Also, we need to support the binary version `bin/knapsack_pro`, which directly calls runner classes.

ArturT: > Sometimes we are calling it args, sometimes options. I think we should stick with options for this one (vs positional arguments.)

Hmm. I wanted to use a general term.
Perhaps the option name could be used once we know these are actual options. I'm not sure if we need to be that specific right away.

We use the args term in many places, also in [the docs](https://docs.knapsackpro.com/ruby/reference/#command-line-arguments).

image

ArturT:
> Are we also supporting --example and --example-matches with this PR?

Yes.

PR #289, opened 2025-02-19
Improve execution time tracking for RSpec individual test examples.

ArturT:
# Story

https://trello.com/c/cq5i0WDt

# Description

Related to [the RSpec split by examples feature](https://docs.knapsackpro.com/ruby/split-by-test-examples/).

# problem

Someone gives a custom set of paths to the Knapsack Pro using `KNAPSACK_PRO_TEST_FILE_LIST_SOURCE_FILE`.
There could be paths like `a_spec.rb[1:1]`. But the `build_distributions#last` returns tests which are not considered slow. `a_spec.rb` could be considered fast. We measure execution time for the full path `a_spec.rb` instead of `a_spec.rb[1:1]`.

This can lead to misleading data because node 0 could measure `a_spec.rb[1:1]` as `a_spec.rb` taking 1 minute. But node 1 could measure `a_spec.rb[1:2]` as `a_spec.rb` taking 2 seconds.
Now we have two `a_spec.rb` paths with different timings. They are merged on the API side. We may end up with a build with `a_spec.rb` taking 2s while, in fact, it took much longer because it was split by examples between 2 nodes. We expect individual test examples to be measured.

# solution

Individual test examples (`a_spec.rb[1:1]`, `a_spec.rb[1:2]`) are measured from now on.

# Checklist reminder


spec/knapsack_pro/config/env_spec.rb
```diff
@@ -1016,7 +1016,10 @@
   describe '.rspec_split_by_test_examples?' do
     subject { described_class.rspec_split_by_test_examples? }
 
-    after(:each) do
+    before do
```
ArturT, line 1019: fix for flaky test

lib/knapsack_pro/runners/queue/rspec_runner.rb
```diff
@@ -183,6 +183,10 @@ def pull_tests_from_queue(can_initialize_queue: false)
             executed_test_files: @node_test_file_paths
           )
           @node_test_file_paths += test_file_paths
+
+          time_tracker = KnapsackPro::Formatters::TimeTrackerFetcher.call
+          time_tracker.scheduled_test_file_paths = @node_test_file_paths
```
3v0k4, line 188: Each time we set what was there before plus the new batch?

spec/integration/runners/queue/rspec_runner_spec.rb
```diff
@@ -2427,6 +2427,104 @@ def when_first_matching_example_defined(type:)
...
+      ENV['KNAPSACK_PRO_SLOW_TEST_FILE_PATTERN'] = ""
+      ENV['KNAPSACK_PRO_CI_NODE_TOTAL'] = '2'
+    end
+    after do
+      ENV.delete('KNAPSACK_PRO_RSPEC_SPLIT_BY_TEST_EXAMPLES')
+      ENV.delete('KNAPSACK_PRO_SLOW_TEST_FILE_PATTERN')
+      ENV.delete('KNAPSACK_PRO_CI_NODE_TOTAL')
+    end
+
+    it 'detects test execution times correctly for individual test examples even though they are not considered as slow test files' do
```
3v0k4, line 2442:
```suggestion
    it 'detects test execution times correctly for individual test examples even though they are not considered slow test files' do
```

spec/integration/runners/queue/rspec_runner_spec.rb
```diff
@@ -2427,6 +2427,104 @@ def when_first_matching_example_defined(type:)
...
+          C2 test example
+        OUTPUT
+      )
+
+      expect(actual.stdout.scan(/A1 test example/).size).to eq 1
+      expect(actual.stdout.scan(/A2 test example/).size).to eq 1
+
+      expect(actual.stdout).to include('6 examples, 0 failures')
+
+      expect(actual.stdout).to include('[INTEGRATION TEST] test_files: 4, test files have execution time: true')
```
3v0k4, line 2522: If the bug was not solved this assertion would fail because in the logs it would be `'[INTEGRATION TEST] test_files: 4, test files have execution time: false'`?

lib/knapsack_pro/formatters/time_tracker.rb
3v0k4, line 1: plea

Let's not add `_test_file_paths` to namings (only `_paths`). It's clear from the context.

Othewise, when I read `@paths` and `@scheduled_test_file_paths` then I'm wondering if `@paths` are tests since it's not called `@test_file_paths`.

lib/knapsack_pro/formatters/time_tracker.rb
```diff
@@ -22,6 +23,13 @@ def initialize(_output)
         @group = {}
         @paths = {}
         @suite_started = now
+        @scheduled_test_file_paths = []
+        @split_by_example_test_file_paths = Set.new
+      end
+
+      def scheduled_test_file_paths=(scheduled_test_file_paths)
+        @scheduled_test_file_paths = scheduled_test_file_paths
+        upsert_split_by_example_test_file_paths
```
3v0k4, line 32: recommendation

I'd inline the upsert. It's more clear to see it in one place.

lib/knapsack_pro/formatters/time_tracker.rb
3v0k4: plea

We use many terms to identify paths with ids or not.

I wouldn't refactor the entire codebase, but let's try to stick to this file's already existing conventions.

I think this was the rule:
- file path (path without id)
- id path (path with id)
- path (either one)

3v0k4 (commented):
I see that to keep things short you used "split by examples" or `split_by_examples`. I'd keep it consistent to not add another way of calling it, so: "split by test examples" or `split_by_test_examples`. If it's a throwaway variable inside a method I'm fine with a shortcut.

Example:

```
# ok (because if I grepped I would find the split line)

def do
  split = Env.split_by_test_examples_enabled?
  # ...
end

# not ok

@split
it "split by example"
```

---

Could you ping me after you apply the changes to take a second look please?

lib/knapsack_pro/adapters/rspec_adapter.rb
```diff
@@ -117,10 +124,19 @@ def log_tests_duration
         ::RSpec.configure do |config|
           config.append_after(:suite) do
             time_tracker = KnapsackPro::Formatters::TimeTrackerFetcher.call
-            if time_tracker
-              formatted = KnapsackPro::Presenter.global_time(time_tracker.duration)
-              KnapsackPro.logger.debug(formatted)
-            end
+            formatted = KnapsackPro::Presenter.global_time(time_tracker.duration)
```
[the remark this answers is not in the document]
ArturT: The `time_tracker` was fetched in the context of `after_subset_queue` hook before:

```ruby
  def log_batch_duration
        KnapsackPro::Hooks::Queue.after_subset_queue do
          time_tracker = KnapsackPro::Formatters::TimeTrackerFetcher.call
          if time_tracker
            formatted = KnapsackPro::Presenter.global_time(time_tracker.batch_duration)
            KnapsackPro.logger.debug(formatted)
          end
        end
      end
```

The ` if time_tracker` condition was added to fix some kind of test that was flaky. Not sure which one exactly.
https://github.com/KnapsackPro/knapsack_pro-ruby/pull/237/commits/dcb6afb5e963275be308b5e2e0da6d03416f8180

I think an `Exception` could have happened, and RSpec never started, so `time_tracker` was not set.
Hence we needed `if time_tracker`.
The `after_subset_queue` hook could be called even when RSpec failed (this was before 7.0 version). It's  `KnapsackPro::Hooks::Queue.call_after_queue` right before `Kernel.exit(1)`.

```ruby
# lib/knapsack_pro/runners/queue/rspec_runner.rb
 begin
              exit_code = rspec_runner.run($stderr, $stdout)
              exitstatus = exit_code if exit_code != 0
            rescue Exception => exception
              KnapsackPro.logger.error("Having exception when running RSpec: #{exception.inspect}")
              KnapsackPro::Formatters::RSpecQueueSummaryFormatter.print_exit_summary(all_test_file_paths)
              KnapsackPro::Hooks::Queue.call_after_subset_queue
              KnapsackPro::Hooks::Queue.call_after_queue
              Kernel.exit(1)
              raise
            else
              if rspec_runner.world.wants_to_quit
                KnapsackPro.logger.warn('RSpec wants to quit.')
                set_terminate_process
              end
              if rspec_runner.world.respond_to?(:rspec_is_quitting) && rspec_runner.world.rspec_is_quitting
                KnapsackPro.logger.warn('RSpec is quitting.')
                set_terminate_process
              end

              printable_args = args_with_seed_option_added_when_viable(args, rspec_runner)
              log_rspec_command(printable_args, test_file_paths, :subset_queue)

              rspec_clear_examples

              KnapsackPro::Hooks::Queue.call_after_subset_queue
```

Source of the PR:
https://github.com/KnapsackPro/knapsack_pro-ruby/pull/237/files

lib/knapsack_pro/adapters/rspec_adapter.rb
```diff
@@ -117,10 +124,19 @@ def log_tests_duration
...
+            KnapsackPro.logger.debug(formatted)
+          end
+        end
+      end
+
+      def bind_regular_mode_time_tracker
+        return unless KnapsackPro::Config::Env.recording_enabled?
```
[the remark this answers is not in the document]
ArturT: `KnapsackPro::Config::Env.recording_enabled?` is true when you use regular mode only.

lib/knapsack_pro/adapters/rspec_adapter.rb
```diff
@@ -117,10 +124,19 @@ def log_tests_duration
...
+
+        ::RSpec.configure do |config|
+          config.append_before(:suite) do
+            time_tracker = KnapsackPro::Formatters::TimeTrackerFetcher.call
```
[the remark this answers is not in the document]
ArturT: Yes. If `before(:suite)` is called, then RSpec is already loaded, which means the TimeTracker formatter is also.

lib/knapsack_pro/runners/queue/rspec_runner.rb
ArturT: Yes.

Each time we override the time_tracker.scheduled_test_file_paths with @node_test_file_paths.

The @node_test_file_paths is updated on each batch pulled from the Queue API.

spec/integration/runners/queue/rspec_runner_spec.rb
ArturT: Yes.

It would be:

```
'[INTEGRATION TEST] test_files: 5, test files have execution time: false'
```

5 test files would be measured in the old approach:

```
a_spec.rb with time
a_spec.rb[1:1] with 0s
a_spec.rb[1:2] with 0s
b_spec.rb with time
c_spec.rb with time
```

lib/knapsack_pro/adapters/rspec_adapter.rb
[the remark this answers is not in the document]
ArturT: I did a little refactoring because the following methods are a bit confusing:

```
KnapsackPro::Config::Env.recording_enabled? # is regular mode on
KnapsackPro::Config::Env.queue_recording_enabled? # is queue mode on
```

I've renamed them to:

```
KnapsackPro::Config::Env.regular_mode?
KnapsackPro::Config::Env.queue_mode?
```

lib/knapsack_pro/adapters/rspec_adapter.rb
```diff
@@ -150,6 +166,10 @@ def bind_after_queue_hook
         end
       end
 
+      def self.scheduled_test_file_paths
```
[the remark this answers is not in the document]
ArturT: > This is public so that it can be tested?

Yes.

> Maybe let's make it private and call it with send from there?

Ok.

lib/knapsack_pro/adapters/rspec_adapter.rb
```diff
@@ -80,6 +80,12 @@ def self.parse_file_path(id)
         id.match(/\A(.*?)(?:\[([\d\s:,]+)\])?\z/).captures.first
       end
 
+      def self.rspec_id_path?(path)
+        path_with_id_regex = /.+_spec\.rb\[.+\]$/
+
+        path&.match?(path_with_id_regex)
+      end
```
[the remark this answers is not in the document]
ArturT: I applied changes: https://github.com/KnapsackPro/knapsack_pro-ruby/pull/289/commits/868dc17a541e8cc378ecdc2bf0b9cd9355c9886e

lib/knapsack_pro/formatters/time_tracker.rb
ArturT: Applied: https://github.com/KnapsackPro/knapsack_pro-ruby/pull/289/commits/3eab4b0965d26c5f1ba1f3551e319e326bfc1a8f

3v0k4:
@ArturT I think the following would be clearer in the PR description:

# Problem

Someone gives a set of paths to Knapsack Pro using `KNAPSACK_PRO_TEST_FILE_LIST_SOURCE_FILE` including `a_spec.rb[1:1]`.

The `build_distributions#last` endpoint may return `a_spec.rb` (not `a_spec.rb[1:1]`), so the Knapsack Pro client measures the following execution times:
- `a_spec.rb` takes as long as `a_spec.rb[1:1]`
- `a_spec.rb[1:1]` takes 0 seconds (default time)

It may happen that:
- node 0 measures `a_spec.rb[1:1]` as `a_spec.rb` taking 1 minute and `a_spec.rb[1:1]` taking 0 seconds
- node 1 measures `a_spec.rb[1:2]` as `a_spec.rb` taking 2 seconds and `a_spec.rb[1:2]` taking 0 seconds

There are two `a_spec.rb` paths with different timings that are merged on the API side. We may end up with a build with `a_spec.rb` taking 2s while, in fact, it took much longer because it was split by examples between 2 nodes. We expect individual test examples to be measured.

Also, the dashboard shows:
- `a_spec.rb`: 2 seconds
- `a_spec.rb[1:1]`: 0 seconds
- `a_spec.rb[1:2]`: 0 seconds

3v0k4 (approved):
Fantastic work!

ArturT:
> @ArturT I think the following would be clearer in the PR description:

No. It's inaccurate.
I'll keep the original description.

> # Problem
> Someone gives a set of paths to Knapsack Pro using `KNAPSACK_PRO_TEST_FILE_LIST_SOURCE_FILE` including `a_spec.rb[1:1]`.
> 
> The `build_distributions#last` endpoint may return `a_spec.rb` (not `a_spec.rb[1:1]`), so the Knapsack Pro client measures the following execution times:
> 
> * `a_spec.rb` takes as long as `a_spec.rb[1:1]`
> * `a_spec.rb[1:1]` takes 0 seconds (default time)
> 

The `build_distributions#last` may return `a_spec.rb` and/or `a_spec.rb[1:1]`. It does not matter. The gem determines the slow test files based on whatever data it gets from the `build_distributions#last` API.

If the full test file path (a_spec.rb) is not determined as slow, but the Queue API returns a_spec.rb[1:1] then we measured execution time for the full path a_spec.rb. We expected a_spec.rb[1:1] to be measured.

lib/knapsack_pro/adapters/rspec_adapter.rb
```diff
@@ -117,10 +123,19 @@ def log_tests_duration
...
+        end
+      end
+
+      def bind_regular_mode_time_tracker
+        return unless KnapsackPro::Config::Env.regular_mode?
+
+        ::RSpec.configure do |config|
+          config.append_before(:suite) do
+            time_tracker = KnapsackPro::Formatters::TimeTrackerFetcher.call
+            time_tracker.scheduled_paths = KnapsackPro::Adapters::RSpecAdapter.scheduled_paths
```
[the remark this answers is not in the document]
ArturT: I'll keep the original code.

`self.class` fails the test:

```
An error occurred in a `before(:suite)` hook.
Failure/Error: super

NoMethodError:
  undefined method `scheduled_paths' for RSpec::Core::AnonymousExampleGroup:Class
# /home/circleci/knapsack_pro-ruby/lib/knapsack_pro/adapters/rspec_adapter.rb:138:in `block (2 levels) in bind_regular_mode_time_tracker'
after suite
```

https://app.circleci.com/pipelines/github/KnapsackPro/knapsack_pro-ruby/1677/workflows/b0882058-be4c-44be-9c8c-0effff1a4405/jobs/26772

3v0k4:
My point was that the description in this PR is confusing, that's why I tried to simplify it.

You may still want to update the description in case someone else gets here and tries to understand what's going on.

PR #292, opened 2025-03-19
Improve the performance of the RSpec split by test examples.

ArturT:
# Story

[link to the internal story](https://trello.com/c/xTeZMWMB)

# Description

Related to the [RSpec split by test examples](https://docs.knapsackpro.com/ruby/split-by-test-examples/).

The following performance improvements were implemented:

* Reduce `/v1/build_distributions/last` API requests.
* (Queue Mode) Improve the speed of starting tests for CI nodes that are started after the queue was already initialized by another CI node.
* (Regular Mode) Improve the speed of starting tests for CI nodes that are started after the test suite split was already initialized by another CI node.

# Checklist reminder

- [x] You added the changes to the `UNRELEASED` section of the `CHANGELOG.md`, including the needed bump (ie, patch, minor, major)


lib/knapsack_pro/queue_allocator.rb
```diff
@@ -15,61 +15,49 @@ def initialize(args)
 
     def test_file_paths(can_initialize_queue, executed_test_files)
```
ArturT, line 16: @3v0k4 This method gives tests to run to Knapsack. It handles Queue Mode.

I like the structure so far. It has a straight flow and guard clauses.

There is also a file `lib/knapsack_pro/test_suite_builder.rb`, which is something new.

Let's sync on a call to see what we like or not before polishing this more.

PR #296, opened 2025-04-09
Avoid needing a shell for rspec

liath:
# Story

We're migrating some services to distroless and generally prefer to run tests in the final built image but Knapsack requires a shell for this one method. I've monkeypatched my way around it in my repos but thought it's a simple change that others may benefit from so here we are.

# Description

Rather than pass these envvars in the command, which forces `system()` to use a shell, we pass them as a hash in the first param. Now we can run knapsack in a container that lacks a shell.

# Changes

TODO: changes introduced by this PR

# Checklist reminder

- [x] You added the changes to the `UNRELEASED` section of the `CHANGELOG.md`, including the needed bump (ie, patch, minor, major)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+### Unreleased
+
+* (patch) RSpec: allow tests to run in a distroless environment
+
+    https://github.com/KnapsackPro/knapsack_pro-ruby/pull/296
+
 ### 8.1.0
 
 * Improve the performance of the [RSpec split by test examples](https://docs.knapsackpro.com/ruby/split-by-test-examples/).
--- a/lib/knapsack_pro/adapters/rspec_adapter.rb
+++ b/lib/knapsack_pro/adapters/rspec_adapter.rb
@@ -25,12 +25,10 @@ module KnapsackPro
 
         # generate the RSpec JSON report in a separate process to not pollute the RSpec state
         cmd = [
-          'RACK_ENV=test',
-          'RAILS_ENV=test',
           KnapsackPro::Config::Env.rspec_test_example_detector_prefix,
-          'rake knapsack_pro:rspec_test_example_detector',
+          'rake knapsack_pro:rspec_test_example_detector'
         ].join(' ')
-        unless Kernel.system(cmd)
+        unless Kernel.system(ENV.to_h.merge({ 'RACK_ENV' => 'test', 'RAILS_ENV' => 'test' }), cmd)
           raise "Could not generate JSON report for RSpec. Rake task failed when running #{cmd}"
         end
 
--- a/lib/knapsack_pro/adapters/rspec_adapter.rb
+++ b/lib/knapsack_pro/adapters/rspec_adapter.rb
@@ -28,7 +28,7 @@ module KnapsackPro
           KnapsackPro::Config::Env.rspec_test_example_detector_prefix,
           'rake knapsack_pro:rspec_test_example_detector'
         ].join(' ')
-        unless Kernel.system(ENV.to_h.merge({ 'RACK_ENV' => 'test', 'RAILS_ENV' => 'test' }), cmd)
+        unless Kernel.system({ 'RACK_ENV' => 'test', 'RAILS_ENV' => 'test' }, cmd)
           raise "Could not generate JSON report for RSpec. Rake task failed when running #{cmd}"
         end
 
--- a/lib/knapsack_pro/repository_adapters/git_adapter.rb
+++ b/lib/knapsack_pro/repository_adapters/git_adapter.rb
@@ -44,26 +44,37 @@ module KnapsackPro
 
       def git_commit_authors
         if KnapsackPro::Config::Env.ci? && shallow_repository?
-          command = 'git fetch --shallow-since "one month ago" --quiet 2>/dev/null'
+          command = 'git fetch --shallow-since "one month ago" --quiet'
           begin
             Timeout.timeout(5) do
-              `#{command}`
+              Kernel.system(command, err: File::NULL)
             end
           rescue Timeout::Error
             KnapsackPro.logger.debug("Skip the `#{command}` command because it took too long.")
           end
         end
 
-        `git log --since "one month ago" 2>/dev/null | git shortlog --summary --email 2>/dev/null`
+        summary_read, log_write = IO.pipe
+        Kernel.system('git log --since "one month ago"', out: log_write, err: File::NULL)
+        log_write.close
+        summary, summary_write = IO.pipe
+        Kernel.system('git shortlog --summary --email', in: summary_read, out: summary_write, err: File::NULL)
+        summary_write.close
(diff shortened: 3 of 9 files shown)


ArturT:
@liath Could you elaborate on the issue that you were experiencing before this change?

Something I'd like to ensure is that we are isolating the command `Kernel.system({ 'RACK_ENV' => 'test', 'RAILS_ENV' => 'test' }, cmd)`, and it won't pollute the memory of the main process that invoked it. Is that still the case?

Thanks.

ArturT:
@liath, could you create a new empty commit so that we can trigger the CI pipeline and verify if the tests are passing? Additional tests run on CI. Thanks.

```
git commit --allow-empty -m "empty commit"
```

liath:
@ArturT Here's the smallest slice of our setup I could make to reproduce the issue, it's still kinda large, sorry 😓 

Save this as `Dockerfile`:
```dockerfile
FROM cgr.dev/chainguard/ruby:latest-dev AS builder

WORKDIR /build
RUN <<RUN
cat <<GEMFILE > Gemfile
source "https://rubygems.org"
gem "knapsack_pro"
gem "logger"
gem "rspec"
GEMFILE

gem install bundler
bundle config set --local path vendor/bundle
bundle install
RUN

RUN <<RUN
cat <<RAKEFILE > Rakefile
require 'knapsack_pro'
KnapsackPro.load_tasks if defined?(KnapsackPro)
RAKEFILE

mkdir bin
cat <<CI > bin/ci
ENV['PATH'] =+ "#{Gem.user_dir}/bin"
require 'bundler/setup'
require 'knapsack_pro'

KnapsackPro::Runners::Queue::RSpecRunner.run('--format d')
CI

mkdir spec
cat <<SPEC_HELPER > spec/spec_helper.rb
require 'knapsack_pro'
KnapsackPro::Adapters::RSpecAdapter.bind
SPEC_HELPER

cat <<TEST_SPEC > spec/test_spec.rb
require 'spec_helper'
RSpec.describe Object do
  it 'meows' do
    expect('meow').to eq('meow')
  end
end
TEST_SPEC
RUN

FROM cgr.dev/chainguard/ruby:latest

WORKDIR /app
COPY --from=builder /home/nonroot/.local /home/nonroot/.local
COPY --from=builder /build /app
CMD ["bin/ci"]

ENV KNAPSACK_PRO_BRANCH=main
ENV KNAPSACK_PRO_CI_NODE_BUILD_ID=0
ENV KNAPSACK_PRO_CI_NODE_INDEX=0
ENV KNAPSACK_PRO_CI_NODE_TOTAL=2
ENV KNAPSACK_PRO_COMMIT_HASH=0000
ENV KNAPSACK_PRO_FIXED_QUEUE_SPLIT=true
ENV KNAPSACK_PRO_RSPEC_SPLIT_BY_TEST_EXAMPLES=true
ENV KNAPSACK_PRO_SLOW_TEST_FILE_PATTERN="{spec/*}"
ENV KNAPSACK_PRO_SLOW_TEST_FILE_THRESHOLD=1
```

Then run:
```shell
# you'll need to export the creds for Knapsack
docker build -t test . && \
  docker run --rm -it -e KNAPSACK_PRO_TEST_SUITE_TOKEN_RSPEC -e KNAPSACK_PRO_SALT -e CI=true test
```

You'll get a fairly opaque stack trace like:
```
 Exception backtrace: /app/vendor/bundle/ruby/3.4.0/gems/knapsack_pro-8.1.1/lib/knapsack_pro/adapters/rspec_adapter.rb:34:in 'KnapsackPro::Adapters::RSpecAdapter.test_file_cases_for'
...
```

We can monkeypatch in my changes from this PR by inserting the following into the bin/ci script before the last line:
```ruby
KnapsackPro::Adapters::RSpecAdapter.class_eval do
  def self.test_file_cases_for(slow_test_files)
    cmd = [
      KnapsackPro::Config::Env.rspec_test_example_detector_prefix,
      'rake knapsack_pro:rspec_test_example_detector'
    ].join(' ')
    unless Kernel.system({ 'RACK_ENV'=>'test', 'RAILS_ENV'=>'test' }, cmd)
      raise "Could not generate JSON report for RSpec. Rake task failed when running #{cmd}"
    end

    # read the JSON report
    KnapsackPro::TestCaseDetectors::RSpecTestExampleDetector.new.test_file_example_paths
  end
end
```

And the test will now pass. 

As for polluting memory space, resource cost should actually be lower as we now aren't spawning a child shell to run the same task. I don't consider myself an expert in ruby however so perhaps there's something non-obvious about `system` that I'm unaware of?

ArturT:
@liath Thank you. I'll bring this to our team.

liath:
This is incomplete fwiw but was the minimum to get us moving. There's [at least a few places](https://github.com/KnapsackPro/knapsack_pro-ruby/blob/7ec77c84da3bd84c6ea2865278a6d68b103b40b6/lib/knapsack_pro/repository_adapters/git_adapter.rb#L47) where we'd need to move shell redirection syntax to the redirection params. Eg:

```ruby
# before:
`git log --format="%aN <%aE>" -1 2>/dev/null`
# after:
system('git log --format="%aN <%aE>" -1', err: File::NULL)
```

ArturT:
@liath Thanks, that's a good point.

I’d like to propose a few alternatives:

1. We could release just the change you committed in this PR.
2. If you’d like to contribute other changes (e.g., `git`), we’d be happy to review and release them too.
3. We could look into the changes required for 2. ourselves, but we may need a bit of time.

What do you think?

liath:
> 2. If you’d like to contribute other changes (e.g., `git`), we’d be happy to review and release them too.

I think I have all of the bases covered for this now. The syntax when the commands output is desired is... verbose, especially around piping: https://github.com/KnapsackPro/knapsack_pro-ruby/pull/296/files#diff-b2689facc4577d4d432168b99958c787d8e07fe8df6891a8dbd2160b9f961751R57

But it should all work as before so the e2e tests failing is weird, I don't think I will have changed error handling anywhere that would cause this. The places I switched backticks to system will hide errors but that's limited to the git adapter. :\

ArturT:
@liath Thank you for your contribution. We're going to review this. It has been added to our backlog. 

Regarding failing e2e tests, CI secrets are not available for builds from external contributors, which makes the tests fail. I'll need to check this.

May we push changes on top of your commits to adjust things and ensure CI works? 
Or we will create a separate PR based on your changes when we review this PR.

Thanks again for your help!

liath:
> Regarding failing e2e tests, CI secrets are not available for builds from external contributors
@ArturT ah, of course lol. I just assumed it was something I broke somehow :P

> May we push changes on top of your commits to adjust things and ensure CI works? 
Yeah, no worries!

ArturT:
@liath We've released a new gem version 8.2.0 that incorporates your changes from this PR. I had to create a separate PR based on your changes here https://github.com/KnapsackPro/knapsack_pro-ruby/pull/299 to run tests on CI.

Thank you for your contribution!

Issue #269, opened 2024-08-15
Infinite recursion when KNAPSACK_PRO_LOG_LEVEL is set to debug


dmytro-workday:
Hi,

We have set `KNAPSACK_PRO_LOG_LEVEL=debug` and `KNAPSACK_PRO_LOG_DIR=log`, which leads to the following backtrace:

```
[2024-08-15 12:56:16] [2024-08-15 12:56:16] [NODE=4] SystemStackError: stack level too deep (SystemStackError)
[2024-08-15 12:56:16] [2024-08-15 12:56:16] [NODE=4] /builds/app/vendor/bundle/ruby/3.2.0/gems/knapsack_pro-7.8.0/lib/knapsack_pro/config/env.rb:240:in `map'
[2024-08-15 12:56:16] [2024-08-15 12:56:16] [NODE=4] /builds/app/vendor/bundle/ruby/3.2.0/gems/knapsack_pro-7.8.0/lib/knapsack_pro/config/env.rb:240:in `detected_ci'
[2024-08-15 12:56:16] [2024-08-15 12:56:16] [NODE=4] /builds/app/vendor/bundle/ruby/3.2.0/gems/knapsack_pro-7.8.0/lib/knapsack_pro/config/env.rb:292:in `ci_env_for'
[2024-08-15 12:56:16] [2024-08-15 12:56:16] [NODE=4] /builds/app/vendor/bundle/ruby/3.2.0/gems/knapsack_pro-7.8.0/lib/knapsack_pro/config/env.rb:282:in `env_for'
[2024-08-15 12:56:16] [2024-08-15 12:56:16] [NODE=4] /builds/app/vendor/bundle/ruby/3.2.0/gems/knapsack_pro-7.8.0/lib/knapsack_pro/config/env.rb:20:in `ci_node_index'
[2024-08-15 12:56:16] [2024-08-15 12:56:16] [NODE=4] /builds/app/vendor/bundle/ruby/3.2.0/gems/knapsack_pro-7.8.0/lib/knapsack_pro.rb:99:in `logger'
[2024-08-15 12:56:16] [2024-08-15 12:56:16] [NODE=4] /builds/app/vendor/bundle/ruby/3.2.0/gems/knapsack_pro-7.8.0/lib/knapsack_pro/config/env.rb:285:in `env_for'
[...]

PR #304, opened 2025-07-10
Fix infinite recursion when logger gets into the loop

ArturT:
# Story

https://trello.com/c/yROeD8FX

## Related

* https://github.com/KnapsackPro/knapsack_pro-ruby/issues/269

# Description

Fix infinite recursion when the logger gets into the loop.

# Changes:

* Log info message only once to prevent infinite recursion
* Do not log the message for CI node index and CI node total conflict if Knapsack Pro is running within the parallel_tests gem context. This message could be confusing as suggested [here](https://github.com/KnapsackPro/knapsack_pro-ruby/issues/269#issuecomment-3056392927).

# Checklist reminder

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -339,6 +339,23 @@ jobs:
             rm .rspec
             mv .rspec.off .rspec
             exit $RSPEC_EXIT_CODE
+      - run:
+          working_directory: ~/rails-app-with-knapsack_pro
+          # https://github.com/KnapsackPro/knapsack_pro-ruby/issues/269
+          command: |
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
 
 ### UNRELEASED (patch)
 
+### 8.3.2
+
+* Fix infinite recursion when the logger gets into a loop
```
3v0k4, line 7: plea

Please add more information here. The issue is a good starting point.

3v0k4 (approved):

PR #310, opened 2025-11-26
Allow to precalculate Split by Test Examples

3v0k4:
# Description

Documentation:

## `KNAPSACK_PRO_RSPEC_SPLIT_BY_TEST_EXAMPLES_FILE` (RSpec)

By default, Knapsack Pro calculates the [Split by Test Examples](split-by-test-examples.mdx) as part of the test run.

You may want to precalculate the split *once* to avoid doing it later in each node running Knapsack Pro. 

To enable the precalculation set `KNAPSACK_PRO_RSPEC_SPLIT_BY_TEST_EXAMPLES_FILE` for both the `bundle exec rake knapsack_pro:rspec:precalculate_split_by_test_examples` and `bundle exec rake knapsack_pro:queue:rspec`.

Default: unset

Example:

```bash
export KNAPSACK_PRO_RSPEC_SPLIT_BY_TEST_EXAMPLES_FILE=report.json

bundle exec rake knapsack_pro:rspec:precalculate_split_by_test_examples

# On each parallel CI node:
bundle exec rake knapsack_pro:queue:rspec
```

# Checks

- [ ] I added the changes to the `UNRELEASED` section of the `CHANGELOG.md`, including the needed bump (i.e., patch, minor, major)
- [ ] I followed the architecture outlined below for RSpec in Queue Mode:
  - Pure: `lib/knapsack_pro/pure/queue/rspec_pure.rb` contains pure functions that are unit tested.
  - Extension: `lib/knapsack_pro/extensions/rspec_extension.rb` encapsulates calls to RSpec internals and is integration and E2E tested.
  - Runner: `lib/knapsack_pro/runners/queue/rspec_runner.rb` invokes the pure code and the extension to produce side effects, which are integration and E2E tested.

PR #322, opened 2026-01-07
`rake knapsack_pro:queue:rspec:initialize`

3v0k4:
# Description

This PR takes over #310: instead of precalculating a split, which is an intermediate step in creating the queue, directly create the queue to be consumed later. This comes with the following advantages:
- Hides an implementation detail: the user wants to initialize the queue, not precalculating something for Knapsack Pro
- No need for `KNAPSACK_PRO_RSPEC_SPLIT_BY_TEST_EXAMPLES_FILE`
- Gives us an opportunity to add/remove steps in the queue initialization later (e.g., coordinate with the API the need to run the detector if the split is already balanced by file)
- Allows the user to initialize the queue in one job and fan out to parallel jobs to consume that queue

# Checks


ArturT (approved):

connorshea:
Is the idea here that you can potentially get a bit of a faster boot-up, if you can run this command prior to the start of the test jobs?

3v0k4:
@connorshea most users won't see a difference, that's why it was shipped silently. It was mostly for us to prepare the ground for later work.

PR #325, opened 2026-01-21
RoF

3v0k4:
# Description

See https://github.com/KnapsackPro/knapsack-pro-api/pull/1455

# Checks

- [ ] I added the changes to the `UNRELEASED` section of the `CHANGELOG.md`, including the needed bump (i.e., patch, minor, major)
- [ ] I followed the architecture outlined below for RSpec in Queue Mode:


rails-app-with-knapsack_pro/bin/bin_knapsack_pro_rspec
```diff
@@ -8,5 +8,6 @@ KNAPSACK_PRO_ENDPOINT=http://api.knapsackpro.localhost:3000 \
   KNAPSACK_PRO_PROJECT_DIR=. \
   KNAPSACK_PRO_CI_NODE_TOTAL=${2:-2} \
   KNAPSACK_PRO_CI_NODE_INDEX=${1:-0} \
+  KNAPSACK_PRO_TEST_QUEUE_ID=${3:-$CI_BUILD_ID} \
```
ArturT: question

Is this needed in the regular mode? I think it's not. You could remove it then.

ArturT (approved):
Great work!

lib/knapsack_pro/config/env.rb
```diff
@@ -164,6 +164,10 @@ def fixed_queue_split
           end
         end
 
+        def fixed_queue_split_?
```
ArturT, line 167: question

What's the purpose of this method here? Is it to avoid using the `fixed_queue_split` method that could print logger info a few times when `KNAPSACK_PRO_FIXED_QUEUE_SPLIT` is not set? If   `@fixed_queue_split ||= false` then it would evaluate the `begin` block each time you call the `fixed_queue_split` method.

Perhaps changing `def fixed_queue_split` to the following would be enough and then we don't need ` fixed_queue_split_?`:

```ruby
def fixed_queue_split
  return @fixed_queue_split if defined?(@fixed_queue_split)

  @fixed_queue_split = begin
    env_name = 'KNAPSACK_PRO_FIXED_QUEUE_SPLIT'
    computed = env_for(env_name, :fixed_queue_split).to_s

    if !ENV.key?(env_name)
      KnapsackPro.logger.info("#{env_name} is not set. Using default value: #{computed}. Learn more at #{KnapsackPro::Urls::FIXED_QUEUE_SPLIT}")
    end

    computed
  end
end
```

lib/knapsack_pro/report.rb
```diff
@@ -42,8 +42,7 @@ def self.save_node_queue_to_api(test_files = nil)
       end
 
       if test_files.empty?
-        KnapsackPro.logger.warn("No test files were executed on this CI node.")
-        KnapsackPro.logger.debug("This CI node likely started work late after the test files were already executed by other CI nodes consuming the queue.")
+        KnapsackPro.logger.info("No tests were executed because the test queue is empty.")
```
ArturT, line 45: suggestion

Why is the test queue empty? Let's make it clear to the user.

```suggestion
        KnapsackPro.logger.info("No tests were executed because the test queue has already been consumed.")
```

spec/knapsack_pro/crypto/encryptor_spec.rb
```diff
@@ -55,4 +55,40 @@
...
+    context 'when test files encrypted flag enabled' do
+      let(:test_files_encrypted?) { true }
+      # let(:encryptor) { instance_double(described_class) }
+
+      before do
+        expect(KnapsackPro::Config::Env).to receive(:salt).at_least(1).and_return('123')
+      end
+
+      it do
+        # expect(described_class).to receive(:new).with(test_files).and_return(encryptor)
```
ArturT: Perhaps comments could be removed here and in L70?

ArturT (approved):

lib/knapsack_pro/report.rb
3v0k4: I'm leaving as is because with RoF, a node may not run anything and it's because previously no tests failed. And "has already been consumed" may be confusing.

lib/knapsack_pro/config/env.rb
3v0k4: It's because we should not print 
> KNAPSACK_PRO_FIXED_QUEUE_SPLIT is not set. Using default value: ...

since we don't allow for that ENV to be configured anymore.

But, nice catch on the wrong memoization (false), fixed.k

ArturT (approved):

PR #343, opened 2026-07-25
Accumulate TimeTracker paths in-place instead of rebuilding the hash

connorshea:
# Description

AI Disclosure: This change was generated using Claude Code, with Opus 5. I have manually reviewed and tested the change for correctness and to ensure it makes sense.

The `example_group_finished` method in the TimeTracker formatter used `Hash#merge` to add the finished test's paths into the accumulator, which meant building a whole new hash. This is run once per top-level group in RSpec, and so the work was being done repeatedly and scaled with the number of top-level `describe` blocks as well as how many paths have already been added to the hash. This PR modifies the accumulator in-place instead to avoid adding memory pressure / triggering GC.

I don't believe this change violates the architecture outlined in the PR template in any way (or at least it doesn't _newly_ violate it, as the formatter already mixed purity and side-effects before this change).

These benchmarks were run on my M5 Pro Mac 24GB RAM w/ Ruby 3.4.9.

**Per-file keys (if not splitting by example), 4 examples per file:**

| files | old | new | speedup | old malloc churn | new |
|---|---|---|---|---|---|
| 500 | 4.2 ms | 3.8 ms | 1.1× | — | — |
| 1000 | 9.2 ms | 7.5 ms | 1.2× | — | — |
| 2000 | 20.9 ms | 15.1 ms | 1.4× | 74 MB | 2.0 MB |
| 4000 | 50.5 ms | 30.3 ms | 1.7× | — | — |

**Split-by-test-example (`@paths` keyed per example — the worst case, but only really gets hit if there are a lot of slow tests):**

| workload | entries | old | new | old malloc | new |
|---|---|---|---|---|---|
| 500 × 20 | 10k | 32.4 ms | 23.0 ms | 103 MB | 4.3 MB |
| 1000 × 20 | 20k | 74.1 ms | 45.6 ms | 404 MB | 7.6 MB |

# Checks


connorshea:
I think the failures are due to the tests needing to either be updated for simplecov 1.x, or we need to restrict the range of the simplecov gem

3v0k4:

PR #344, opened 2026-07-25
perf: resolve RSpec test file paths without allocating per example

connorshea:
# Description

AI Disclosure: This was generated using Claude Code with Opus 5. It has been reviewed and tested by me manually.

`RSpecAdapter.file_path_for` runs on every test example (in Queue Mode, it runs twice per example) and was allocating more objects than necessary to check candidate paths. We can instead check it inline while still keeping the laziness.

We built a MatchData and a few other things for every call of `parse_file_path` and `id_path?`. We don't need to do that here, because we can instead look for the last `[` in the path and then match on the string from there onward, without needing to allocate almost anything.

`parse_file_path` now finds `[` with `rindex` and validates it with an allocation-free `Regexp#match?` starting at that offset, and `id_path?` is now an `end_with?` + `match?`.

The only functional change in this code is that a path with a linebreak (which shouldn't happen anyway) would previously cause a NoMethodError, and is now ignored/kept instead. I don't think that's a problem because it doesn't ever happen, but figured I'd note it.

Also wanted to note that I had Claude fuzz the new regex to ensure equivalent behavior on 250k values (some generated to match real paths, some fully random values).

---

This saves a tiny bit of time, but the main benefit of this change is reducing the # of object allocations because it helps avoid triggering GC in large suites, which speeds things up. This isn't going to be hugely significant on its own, but I think it's worth doing.

**Per-call, Ruby 3.4.9, arm64-darwin25, 500 RSpec example ids, `benchmark-ips` (3s, 1s warmup)**

| Method | Objects before | Objects after | µs before | µs after | Speedup |
| --- | ---: | ---: | ---: | ---: | ---: |
| `file_path_for` | 13 | 1 | 0.926 | 0.206 | 4.5x |
| `parse_file_path` (id path) | 6 | 1 | 0.531 | 0.153 | 3.5x |
| `parse_file_path` (plain path) | 5 | 0 | 0.493 | 0.040 | 12.4x |
| `id_path?` (id path) | 6 | 0 | 0.541 | 0.089 | 6.1x |
| `id_path?` (plain path) | 5 | 0 | 0.511 | 0.038 | 13.5x |

**Scaled to a suite (one `file_path_for` per example, NOT Queue Mode)**

| Examples | Time before | Time after | Objects before | Objects after |
| ---: | ---: | ---: | ---: | ---: |
| 10,000 | 9.1 ms | 2.0 ms | 130,007 | 10,007 |
| 100,000 | 94.6 ms | 20.6 ms | 1,300,000 | 100,000 |

**Scaled to a suite (Queue Mode: two `file_path_for` calls per example)**

| Examples | Time before | Time after | Objects before | Objects after |
| ---: | ---: | ---: | ---: | ---: |
| 10,000 | 18.2 ms | 4.0 ms | 260,014 | 20,014 |
| 100,000 | 189.2 ms | 41.2 ms | 2,600,000 | 200,000 |

# Checks


3v0k4:

connorshea:
> Hey @connorshea, thanks for the PR!
> 
> Quick question to help us prioritize the work:
> 
> Is this creating problems in your projects and requires a quick release?

Same answer as the other PRs, not a blocker for us currently, just trying to make knapsack a bit more efficient to speed up our CI pipelines, but no rush on these changes :)

PR #346, opened 2026-07-25
Build `before`/`after(:all)` prefixes once per group in the `TimeTracker`

connorshea:
# Description

AI Disclosure: This was generated using Claude Code with Opus 5. It has been tested and reviewed by me.

`add_hooks_time` re-derived the same `group_id_path[0..-2]` string for every `(example, group)` pair, and `reduce` over a hash allocates an array per pair. This PR builds the prefixes once per group and sums with `each`.

For 500 test files with 40 examples and 8 groups each: 356k allocations / 20.7 ms -> 9k / 8.8 ms.

Though it should be noted that this only happens when the split-by-example mode is enabled and actually gets used (when the Knapsack API marks a file as slow), and the benchmark is inherently unrealistic to show the scale of the change. In most test suites, this likely will not have much or any impact. If this change doesn't feel worth it due to the lack of impact for most users, it'd be fine with me to close it.

# Checks


3v0k4:

PR #347, opened 2026-07-25
Fix failing rspec test by reading coverage data from `coverage_data.js`

connorshea:
# Description

AI Disclosure: Had Claude investigate and fix the problem, verified it myself for correctness.

Fix the test failure caused by the simplecov version upgrade.

SimpleCov 1.0 rewrote its HTML formatter: `index.html` is now static and loads a client-side viewer, and the report data lives in `coverage_data.js`. So we want to check that file instead.

This is a test change so I don't think it makes sense to update the CHANGELOG?

# Checks


3v0k4:
Thanks, @connorshea 🙏

PR #348, opened 2026-07-30
Cucumber: add support for Split by Test Examples (split slow feature files by scenario)

alexspeller:
## Summary

This PR adds [Split by Test Examples](https://docs.knapsackpro.com/ruby/split-by-test-examples/) support for Cucumber, so slow `.feature` files can be split by scenario across parallel CI nodes — mirroring the existing RSpec feature. The docs currently say "Only RSpec is supported. Let us know if you use a different test runner" — we do! 🙂

It is opt-in via `KNAPSACK_PRO_CUCUMBER_SPLIT_BY_TEST_EXAMPLES=true` and requires Cucumber >= 4.0.

## Why this is a natural fit for Cucumber

Cucumber scenarios have a native command-line address: `features/a.feature:12`. Unlike RSpec example ids (`spec/a_spec.rb[1:2]`), scenario line numbers are stable file properties, and Cucumber has supported running `path:line` (including individual Scenario Outline example rows) since forever. I verified the underlying mechanics empirically against Cucumber 4.1, 7.1, and 11.1:

- `cucumber --dry-run --format json --out report.json <files>` lists every runnable scenario as an element with `uri` + `line`, with Scenario Outline example rows expanded to one element per row — on all three versions, exit code 0.
- Running `cucumber features/a.feature:15` (an outline example row line) runs exactly that pickle, and `test_case.location.file` / `.line` in an `Around` hook match the dry-run values — so times can be tracked per scenario path.
- A `file:line` that a `--tags` filter excludes exits 0 (even with `--strict`), so stale schedules can't fail a build.

## How it works

The generic split machinery already flows through `KnapsackPro::TestSuite#calculate_test_files` → `adapter_class.split_by_test_cases_enabled?` / `calculate_slow_id_paths` / `concat_test_files` for both Regular Mode and Queue Mode, so this PR implements the adapter contract for Cucumber and mirrors the RSpec architecture piece by piece:

- **`CucumberAdapter.split_by_test_cases_enabled?`** — gated on the new env var, raises for Cucumber < 4.0 (analogous to the RSpec >= 3.3.0 gate).
- **`knapsack_pro:cucumber_test_example_detector` rake task** (shelled out like the RSpec detector, to not pollute the runner process) fetches slow test files from the API build distribution, then runs an in-process `Cucumber::Cli::Main` dry run with `--format json` to enumerate scenario paths. The user's cucumber args are forwarded (parsed with `Shellwords` so `--tags "not @slow"` survives), so tag filters are respected during scenario enumeration. The task clears `KNAPSACK_PRO_REGULAR_MODE_ENABLED` / `KNAPSACK_PRO_QUEUE_MODE_ENABLED` before the dry run so the adapter bound in `features/support/env.rb` does not register time tracking or save an (empty) report to the API from inside the detector process.
- **`TestCaseMergers::CucumberMerger`** merges `file:line` entries from the API back into whole-file times before slow-file determination (same max-of-file-vs-sum semantics as the RSpec merger).
- **Time tracking:** the `Around` hook now records time under `features/a.feature:12` when that exact test example path is scheduled on the node (new `Tracker#scheduled_test_path?`, which reuses the existing prerun-tests report on disk to work across processes for Regular Mode's forked rake task and Queue Mode's per-batch cucumber processes), and under the file path otherwise.
- Regular and Queue cucumber runners export `KNAPSACK_PRO_CUCUMBER_OPTIONS` for the detector, mirroring `KNAPSACK_PRO_RSPEC_OPTIONS`.

No behavior changes when the feature flag is off (the default).

## Testing

- Unit specs for all new code paths, mirroring the existing RSpec specs (detector, merger, adapter methods, env config, tracker, runners). Full suite: 993 examples, 0 failures.
- End-to-end sanity checks against a real cucumber project (Cucumber 11.1, plus dry-run/`file:line`/`Around`-hook mechanics also verified on 4.1 and 7.1):
  - `TestSuite#calculate_test_files` with the feature enabled replaces a slow feature file with its scenario paths (tag-filtered scenarios excluded) while keeping fast files whole:
    ```
    features/b.feature
    features/a.feature:3
    features/a.feature:6
    features/a.feature:14
    features/a.feature:15
    ```
  - Running cucumber with those paths records per-scenario times for the split file (including an outline example row) and per-file time for the unsplit file.

## Notes for reviewers

- I could not test the API server side, since it is closed source. The client-side merger treats `file:line` as opaque path strings exactly the way `spec/a_spec.rb[1:2]` id paths are treated, so I *believe* no API changes are needed — but you would know best whether anything server-side special-cases path formats.
- The env var defaults to `false` (opt-in), unlike the RSpec one which now defaults to `true` — that seemed the safe choice for a new feature; happy to change.
- Spinach is untouched.


alexspeller:
Note: converting to draft whilst we test drive this against our cucumber suite for a while, will promote again once we're confident in it

alexspeller:
Pushed a second commit (bb8e507) adding an **experimental Queue Mode preload mode for Cucumber** (`KNAPSACK_PRO_CUCUMBER_QUEUE_PRELOAD=true`), because a real-world trial of scenario splitting surfaced a structural cost in Cucumber Queue Mode.

**Trial data** (Rails app, 12 CI nodes, ~14 min/node baseline): enabling Split by Test Examples split 21 slow feature files into 400+ scenario queue items, which grew batch counts from ~7 to ~50 per node. Because Queue Mode shells out a brand-new `cucumber` process (full Rails boot, ~25s) per batch, nodes went from ~14 min to ~24.5 min — the boot overhead swamped the balancing win.

The preload mode boots Cucumber once per node (a dry run loads the support code), then runs each batch in a forked child that reuses the preloaded `Cucumber::Runtime` via the public `Cucumber::Cli::Main#execute!(existing_runtime)` seam (the old spork/DRb integration point — still present and working in Cucumber 4 through 11). Per-batch reporting, hooks, and exit-status semantics are unchanged: children run the same at_exit reporting as the current per-process model. POSIX-only (raises a clear error without `Process.fork`); opt-in, default off.

One notable detail: `Cucumber::Runtime` memoizes reports/formatters/parsed features against its first configuration, so the child resets those ivars before reconfiguring — otherwise results are published to the stale event bus and the run exits 0 even on failures. Verified against Cucumber 9.2.1 and 11.1.1 (single boot, correct per-scenario times, correct failure exit codes). Happy to split this into its own PR if you'd prefer to review it separately.

alexspeller:
Trial update — the preload mode is now validated end-to-end on a real CI suite (Rails app, 12 nodes, Capybara/Selenium browser tests, ~720 scenarios):

| Configuration | Slowest shard | Result |
|---|---|---|
| Baseline (file-level split, process-per-batch) | ~14m17s | green |
| Scenario split, process-per-batch | ~24m30s+ | 1 shard hit 25m timeout |
| Scenario split + preload mode (`KNAPSACK_PRO_CUCUMBER_QUEUE_PRELOAD=true`) | **18m37s** | all green, 20s spread |

Getting there surfaced three fork-hygiene bugs, all fixed on the branch (commits `5f99f10`, `8d38b22`):

1. **Preload dry run must be minimal.** Dry-running the whole suite cost ~8 min (step matching runs even in dry runs); it now dry-runs a single sample file — the dry run only exists to load the support code.
2. **The parent must stay inert.** cucumber.yml profiles can register formatters; one that touches Capybara boots the Capybara server (and a browser) in the parent, and forked children then reuse the parent's server via Capybara's inherited session/port registries — running requests in a process that can't see the child's open DB transactions, and the parent's lingering browser held the CI step's pipes open until job timeout. The preload dry run now strips `-p/--profile` and passes `--no-profile`, and children clear `Capybara.session_pool` / `Capybara::Server.ports` after fork.
3. **ActiveRecord leases survive fork.** Leases are keyed by thread identity, so a child's first checkout could return the parent's connection object and reconnect it in place (undefined behavior on a forked libpq handle — segfaults in pg on macOS). The parent now releases leases and clears pools after preloading; there's also a new `KnapsackPro::Hooks::Queue.after_preload_fork` hook for app-specific resources (the trial app uses it for `SemanticLogger.reopen`).

The remaining ~4 min over baseline is per-batch browser/Capybara-server startup and the split-detector dry run at queue init — inherent to batch granularity rather than to the preloader. Note the trial deliberately forced aggressive splitting (`KNAPSACK_PRO_SLOW_TEST_FILE_THRESHOLD=120`) to exercise the feature; with default settings this suite splits nothing, and preload mode alone would remove the ~7-batches × ~25s of Rails boots that the process-per-batch design pays today.

alexspeller:
One more data point from the trial suite: **preload mode pays off even with scenario splitting disabled**. With default slow-file settings (nothing splits in this suite) and only `KNAPSACK_PRO_CUCUMBER_QUEUE_PRELOAD=true`:

| Configuration | Slowest shard |
|---|---|
| Stock process-per-batch (baseline) | 14m17s |
| **Preload mode, no scenario splitting** | **12m08s** (all 12 shards 10m27s–12m08s, green) |
| Scenario splitting + preload | 18m37s |
| Scenario splitting, process-per-batch | 24m30s+ (timeout) |

That's ~15% off the wall clock purely from booting Rails once per node instead of once per batch (~7 batches × ~25s here). So the preload commit is useful to any Cucumber Queue Mode user with a slow-booting app, independent of the Split by Test Examples feature — it may be worth splitting into its own PR for that reason. Happy to do that if preferred.

3v0k4:
@alexspeller Thanks for the PR! We'll look at the changes in the next couple of weeks 🙏

I understand in the meanwhile you are not blocked because you are using this code on CI, correct?

alexspeller:
We're experimenting with the preload mode and it's working well - I plan to spend a bit more time on our test suite to figure out if we can lose the capybara penalty and take full advantage of the splitting, so we're not blocked!

